PR #3, opened 2019-02-13
Overwrite modulemap and apinotes symlinks

drodriguez:
Otherwise the build will fail if the symlinks are already there.


compnerd:
I changed this to a `copy /y` already and forgot to pull that over.  Sorry!

PR #54, opened 2019-10-25
Adding code to insert a dump of the substitutions for TestRunner.py

plotfi:


plotfi:
This is temporary. We are trying to diagnose what is going on with the busted azure tests. We were not able to repro with docker locally but Gwen and I looked at the failures in Azure and it looks like there is something wrong with one of the substitutions.

PR #60, opened 2019-11-12
Pass the Python executable found by UsePythonVersion to CMake.

drodriguez:
In Windows, CMake FindPythonInterpr uses the registry to find Python if none if provided. Instead of leaving the decision to chance of which Python has written last into the registry, pass the Python found by UsePythonVersion into CMake.

To allow both Unix and Windows, and pipelines that finds an specific Python or not, the template parameter is optional and expands or not into a CMake parameter. This should keep the working pipelines the same, while modifying the failing pipeline to work.


drodriguez:
Seems that the variables are indeed evaluated after the `UsePythonVersion@0` step. The code is now so much simpler.

PR #70, opened 2019-12-05
icu: build ICU data on Linux

compnerd:


toulouse:
flagging this as the blocking issue to have an end to end working system that I can start writing samples and docs for

PR #73, opened 2019-12-09
utilities: change windows-swift to windows_swift to ease import when debugging

toulouse:


compnerd:
I would say that if `import` support is desirable, then, this should be done as is usually the practice with python packages: create a `windows_swift` module and import that from `windows-swift`.

toulouse:
Support isn't particularly desirable so much as the better name is
convenient in this specific and infrequent but useful case. I could do just
as well symlink or copy windows_swift.py to windows-swift.py by hand to
achieve the same effect, but I'm not looking to turn windows-swift.py into
a support library/module in general as it doesn’t need to be at this time.
Since there were no dependencies on the script, I figured I'd make the
tweak.

Fine with rejecting the PR but it doesn’t seem to make anything
structurally worse, either.

compnerd:
It does make it more annoying to type, which is why I tend to prefer the `-` over the `_` (especially since I don't always have working tab completion).

toulouse:
ok boomer

PR #77, opened 2019-12-10
Provide schedules in cron format in YAML.

drodriguez:
This will need removing the existing schedules in the UI for the VS2017 build.

I don’t know how to test this, but the VS2017 pipeline will use the UI, so that one will not break, and we can see if I screw up the syntax in the VS2019 pipeline (which has no schedules in the UI at the moment).

### DIFF
--- a/.ci/toolchain-vs2017-amd64-facebook.yml
+++ b/.ci/toolchain-vs2017-amd64-facebook.yml
@@ -14,6 +14,19 @@ trigger:
     include:
       - .ci/azure-pipelines-aws-agent.yml
       - .ci/templates/toolchain.yml
+schedules:
+- cron: "0 */3 * * *"
+  displayName: "O'clock builds"
+  branches:
+    include:
+    - master
+  always: true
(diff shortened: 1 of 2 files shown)


.ci/toolchain-vs2017-amd64-facebook.yml
```diff
@@ -14,6 +14,19 @@ trigger:
     include:
       - .ci/azure-pipelines-aws-agent.yml
       - .ci/templates/toolchain.yml
+schedules:
+- cron: "0 */3 * * *"
```
compnerd: Can we not do something like: '*/90 * * * *'?

drodriguez: I don't think that's a valid cron syntax. If you Google for "cron 90 minutes" you will find always solutions with 2 entries.

compnerd: Boo!

compnerd:
Lets see what breaks when we merge this 🤷‍♂

PR #100, opened 2019-12-12
windows: templatize Windows SDK

compnerd:


.ci/templates/windows-sdk.yml
```diff
@@ -1,201 +1,233 @@
...
+          allowPartiallySucceededBuilds: true
+          artifactName: 'curl-windows-${{ parameters.host }}'
+          downloadPath: '$(System.ArtifactsDirectory)'
+        displayName: 'Install CURL'
+      - task: BatchScript@1
+        inputs:
+          filename: C:/Program Files (x86)/Microsoft Visual Studio/${{ parameters.VisualStudio }}/Common7/Tools/VsDevCmd.bat
+          arguments: -no_logo -arch=${{ parameters.host }} -host_arch=x64
+          modifyEnvironment: true
+        condition: eq( variables['Agent.OS'], 'Windows_NT' )
```
drodriguez: Maybe move this `condition` higher in the tree (so the template is only valid for Windows, but I don't know if that can be done), apply it also to the checkout step (which uses `script`, but uses batch syntax anyway), or remove it.

.ci/windows-sdk-facebook-vs2017.yml
```diff
@@ -4,244 +4,58 @@ pr:
...
-  - job: windows_armv7
-    condition: false
-    pool: Facebook-VS2017
-    variables:
+  - template: templates/windows-sdk.yml
+    parameters:
+      VisualStudio: 2017/Community
+      pool: Facebook-VS2017
       branch: master
+      toolchain: 5
```
drodriguez: I am sure everyone might appreciate a textual comment about which toolchain IDs are being used. I have to look in the template how it was expanded, but something like `toolchain_pipeline_id` or just `toolchain_pipeline` might had help.

compnerd:
Yeap, happy to invite people to the slack!

.ci/templates/windows-sdk.yml
compnerd: Yeah, I want to generalize the script further, I would rather leave it here since it took a while to puzzle out how to do this.

.ci/windows-sdk-facebook-vs2017.yml
compnerd: I like the comment idea!  I'll actually add that in a follow up change (the testing cycle here is pretty insane)

PR #126, opened 2019-12-16
[FB] Spread out builds. Move Build Completion trigger to YAML.

drodriguez:
Since there's only one machine in the pools for VS2017 and VS2019, builds every 1h30min were a little bit too much for the machine, since the SDK was also build in the same machine and both of them take around 2 hours, so the builds were starting to queue up.

Move the toolchain builds around 2 hours apart, which should avoid the queueing. Also move the Build Completion triggers to YAML to have them under source control management.


drodriguez:
@compnerd: not completely sure about the 'source' field in the changes, but that was the name in the UI.

PR #166, opened 2020-01-18
Cache CMAKE_Swift_LINK_LIBRARY_FLAG

lxbndr:
`CMAKE_Swift_LINK_LIBRARY_FLAG` is not cached without `CACHE` directive.


compnerd:
Bleh, this shouldn't really break anything I think.

PR #167, opened 2020-01-21
[CURL] correct zlib path

lxbndr:
Hope it works same way as in sdk template where I got the path variable.


compnerd:
Ugh, thanks for finding and fixing that.

compnerd:
It definitely fixes the issue:
```
-- Enabled features: IPv6 libz NTLM
-- Enabled protocols: FTP HTTP SMB
```

PR #168, opened 2020-01-21
Publish non-test SDK components on Windows

### LINKED ISSUE #160: Publish non-test binaries for Windows SDK
We should reorder the build steps for the Windows SDK to publish non-testing enabled binaries.

lxbndr:
An attempt to resolve #160 

Basic ideas:
- CMake cache for Swift builds is modified to allow to specify direct StdLib binaries directory instead of SDK installation path. Existing behavior is preserved (perhaps). This prevents re-build steps from conflicting with already installed binaries.
- Stdlib, libdispatch, Foundation, XCTest are built in a row without testing support, then installed and published.
- libdispatch is reconfigured for testing and tested after publishing.
- As well as Foundation.

Checked this approach locally with custom build script. Let's see how actual pipelines will react.

### DIFF
--- a/cmake/caches/windows-x86_64-swift-flags.cmake
+++ b/cmake/caches/windows-x86_64-swift-flags.cmake
@@ -1,9 +1,9 @@
 set(CMAKE_Swift_COMPILER_TARGET x86_64-unknown-windows-msvc CACHE STRING "")
-set(CMAKE_Swift_FLAGS "-resource-dir \"${CMAKE_Swift_SDK}/usr/lib/swift\" -L${CMAKE_Swift_SDK}/usr/lib/swift/windows" CACHE STRING "")
-set(CMAKE_Swift_LINK_FLAGS "-resource-dir \"${CMAKE_Swift_SDK}/usr/lib/swift\" -L${CMAKE_Swift_SDK}/usr/lib/swift/windows" CACHE STRING "")
 
-set(CMAKE_SWIFT_FLAGS "-resource-dir \"${CMAKE_Swift_SDK}/usr/lib/swift\"" CACHE STRING "")
-set(CMAKE_SWIFT_LINK_FLAGS "-resource-dir \"${CMAKE_Swift_SDK}/usr/lib/swift\"" CACHE STRING "")
+set(SWIFT_STDLIB_DIR "${CMAKE_Swift_SDK}/usr" CACHE STRING "")
+
+set(CMAKE_Swift_FLAGS "-resource-dir \"${SWIFT_STDLIB_DIR}/lib/swift\" -L${SWIFT_STDLIB_DIR}/lib/swift/windows" CACHE STRING "")
+set(CMAKE_Swift_LINK_FLAGS "-resource-dir \"${SWIFT_STDLIB_DIR}/lib/swift\" -L${SWIFT_STDLIB_DIR}/lib/swift/windows" CACHE STRING "")
 
 if(CMAKE_VERSION VERSION_LESS 3.16.0)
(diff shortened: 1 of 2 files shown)


cmake/caches/windows-x86_64-swift-flags.cmake
```diff
@@ -1,9 +1,12 @@
...
 
-set(CMAKE_SWIFT_FLAGS "-resource-dir \"${CMAKE_Swift_SDK}/usr/lib/swift\"" CACHE STRING "")
-set(CMAKE_SWIFT_LINK_FLAGS "-resource-dir \"${CMAKE_Swift_SDK}/usr/lib/swift\"" CACHE STRING "")
+set(SWIFT_STDLIB_DIR "${CMAKE_Swift_SDK}/usr" CACHE STRING "")
+
+set(CMAKE_Swift_FLAGS "-resource-dir \"${SWIFT_STDLIB_DIR}/lib/swift\" -L${SWIFT_STDLIB_DIR}/lib/swift/windows" CACHE STRING "")
+set(CMAKE_Swift_LINK_FLAGS "-resource-dir \"${SWIFT_STDLIB_DIR}/lib/swift\" -L${SWIFT_STDLIB_DIR}/lib/swift/windows" CACHE STRING "")
+
+set(CMAKE_SWIFT_FLAGS "-resource-dir \"${SWIFT_STDLIB_DIR}/lib/swift\"" CACHE STRING "")
+set(CMAKE_SWIFT_LINK_FLAGS "-resource-dir \"${SWIFT_STDLIB_DIR}/lib/swift\"" CACHE STRING "")
```
compnerd: I think that these two can go.

cmake/caches/windows-x86_64-swift-flags.cmake
```diff
@@ -1,9 +1,12 @@
 set(CMAKE_Swift_COMPILER_TARGET x86_64-unknown-windows-msvc CACHE STRING "")
-set(CMAKE_Swift_FLAGS "-resource-dir \"${CMAKE_Swift_SDK}/usr/lib/swift\" -L${CMAKE_Swift_SDK}/usr/lib/swift/windows" CACHE STRING "")
-set(CMAKE_Swift_LINK_FLAGS "-resource-dir \"${CMAKE_Swift_SDK}/usr/lib/swift\" -L${CMAKE_Swift_SDK}/usr/lib/swift/windows" CACHE STRING "")
 
-set(CMAKE_SWIFT_FLAGS "-resource-dir \"${CMAKE_Swift_SDK}/usr/lib/swift\"" CACHE STRING "")
-set(CMAKE_SWIFT_LINK_FLAGS "-resource-dir \"${CMAKE_Swift_SDK}/usr/lib/swift\"" CACHE STRING "")
+set(SWIFT_STDLIB_DIR "${CMAKE_Swift_SDK}/usr" CACHE STRING "")
```
compnerd: I suppose that this is okay ... since it computes it based on the variable that we want users to use longer-term.

lxbndr: Not sure about naming though. Is it ok as is? `CMAKE_Swift_STDLIB` better follows general style, but it is not really related to CMake Swift support.

cmake/caches/windows-x86_64-swift-flags.cmake
lxbndr: Done

cmake/caches/windows-x86_64-swift-flags.cmake
compnerd: Yeah, I expect to make `CMAKE_Swift_SDK` a proper CMake variable, but the handling here is special, and so not using the CMake spelling is fine I think.

PR #170, opened 2020-01-29
Match the repository usage in both stages of the build.

drodriguez:
Seems that in a multi-stage pipeline, the collectionId and the definitionId in each of the stages is shared. Azure uses those to find out which repository to look for a manifest file. When the two stages have different repository lists, it seems that the previous work directory is discarded. Since both stages had different repository lists, one stage was overriding the other all the time. Since GC doesn't happen (because there's no maintenance), the file system was being filled all the time.

In theory.

### DIFF
--- a/.ci/templates/toolchain.yml
+++ b/.ci/templates/toolchain.yml
@@ -100,12 +100,18 @@ jobs:
         condition: eq( variables['Agent.OS'], 'Windows_NT' )
         displayName: Enable symbolic links, disable line ending conversion
 
+      # The checkout list has to match with the windows-sdk.yml checkout list.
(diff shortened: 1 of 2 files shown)


compnerd:
If this fixes it, I’m all for it!

drodriguez:
Seems that works in the pipelines. Both VS2017/9 FB are using the same work directory for the last 4 stages (two different builds) each. If the Flowkey setup is similar, it might need a similar fix to avoid wasting space.

compnerd:
@drodriguez awesome; just updated the resources list to synchronize the repositories across Linux and Windows now.  Hopefully we can now have stability again.

PR #183, opened 2020-03-11
Updated Windows.md (source for installers, arguments for cmake, VS 2019 as default, CMake 3.16+ recommended)

stefanspringer1:


docs/Windows.md
```diff
@@ -31,10 +31,10 @@ The installation instructions that follow will result in a directory tree that h
...
 2. Choose `Pipelines` > `Pipelines` from the left of the dashboard.
-3. Choose 'Runs'.
-4. Use the filter symbol to search for "VS2017".
-5. In the results list, scroll down until you see the first successful build (with a green OK symbol) that does not have "facebook" in the description and click on it.
-6. Under "Artifacts", click on the "... published" link.
+3. Use the filter symbol to search for "VS2019".
+4. Click on the appropriate pipeline (e.g. `VS2019`).
+5. The list of the runs will be displayed, scroll down until you see the first successful build (with a green OK symbol) and click on it.
+6. Click the link under "Artifacts:".
 7. Download windows-toolchain-amd64.msi, windows-sdk.msi, and windows-runtime-amd64.msi by clicking on the appropriate down-arrows on the right. Be sure to really download these files from the same build (i.e. do not switch the build for the next download, and be careful when updating). These files will be downloaded as zip files. Unless they are not automatically unzipped during the download process, unzip them to obtain \*.msi files in the extracted directories.
```
compnerd: I think you’re in a better position than I to judge this: should this just be changed to use the swift-build utility to download instead?

At the very least we should clear up the fact that a failure isn’t always a failure - the VS2019 builds Android too, and that a couple of tests are flakey (like in libdispatch and foundation).  Let’s clear it up to indicate that those are safe to use.

docs/Windows.md
```diff
@@ -61,7 +61,19 @@ You will need the ICU libraries from [ICU - International Components for Unicode
 
 #### Building Swift code
 
-You should use a CMake project to build a Swift program. An example CMake project with support for mixed-language support is available at [https://github.com/compnerd/swift-cmake-demo](https://github.com/compnerd/swift-cmake-demo), sample commands that build the project are documented there. The make commands are to be executed from the Native Tools Command Prompt. Note that the cmake executable to be used is the one mentioned above. (Try `cmake -version` to see which CMake version is actually being called.)
+You should use a CMake project to build a Swift program. As an example CMake project use the "HelloWorld-CMake" example from [https://github.com/compnerd/swift-cmake-demo](https://github.com/compnerd/swift-cmake-demo) and use the following commands from within the project directory to build the project for Windows:
+
+```cmd
```
compnerd: The library is part of the install, so perhaps the installation_dir should be C:?

docs/Windows.md
```diff
@@ -61,7 +61,19 @@ You will need the ICU libraries from [ICU - International Components for Unicode
...
+SET INSTALLATION_DIR=C:\Library
+SET SDK=%INSTALLATION_DIR%\Developer\Platforms\Windows.platform\Developer\SDKs\Windows.sdk
+SET OS=windows
+cmake -G Ninja -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo -DBUILD_TESTING=YES -D CMAKE_Swift_FLAGS="-sdk %SDK% -I %SDK%/usr/lib/swift -L %SDK%/usr/lib/swift/%OS%"
+cd build
+ninja
+ninja test
+```
+
+Here, the variables INSTALLATION_DIR, SDK, and OS are just added for clarity of the subsequent command (as used here, they they should not contain spaces). Execute these commands from the Native Tools Command Prompt (or set needed paths before executing). Note that the cmake executable to be used is the one mentioned above. (Try `cmake -version` to see which CMake version is actually being called.)
```
compnerd: I think that we should be resilient to spaces in the path.  If that causes problems we should address them.

docs/Windows.md
```diff
@@ -53,7 +53,7 @@ The installation of the Swift Runtime adds the following value to the system PAT
 
 #### Further requirements: CMake
 
-On Windows, the most convenient setup for building Swift projects currently involves the use of CMake. This requires CMake 3.15+ for Swift support. You can download CMake from [https://cmake.org](https://cmake.org/).
+On Windows, the most convenient setup for building Swift projects currently involves the use of CMake. This requires CMake 3.15+ for Swift support. CMake 3.16+ is recommended. You can download CMake from [https://cmake.org](https://cmake.org/).
```
compnerd: Good call, 3.16 is nearly a requirement for sanity.  The compiler changes to the parallel compile make 3.15 painful.

docs/Windows.md
stefanspringer1: I will have a look at the swift-build utility, maybe add a 2nd possibility in the description, we should first make the current description work.

docs/Windows.md
stefanspringer1: Done.

docs/Windows.md
stefanspringer1: Using the Windows command prompt here and using CMAKE_Swift_FLAGS with some parameters for the Swift compiler poses a problem when using spaces, and I have not been able to resolve that. The problem does not exist when you use a tool (editor, IDE) where you specify each argument separately:

~~~
"cmake",
...
"-D",
"CMAKE_Swift_FLAGS=-sdk \"" + SDK + "\" -I \"" + SDK + "/usr/lib/swift\" -L \"" + SDK + "/usr/lib/swift/" + OS + "\""
~~~

That works. So maybe we should just leave the command line as it is, I added "...better use a tool where you can define each argument to cmake separately" in the description. (We could make a little program that does that, but maybe this goes too far.)

stefanspringer1: New text changed to "...better use a tool where you can define each argument to cmake separately and then put the paths inside the CMAKE_Swift_FLAGS argument in quotes".

stefanspringer1: ...and added "...instead of the outer quotes".

stefanspringer1: ... "...each path...", not "...the paths...".

stefanspringer1: Oh oh... Incomplete testing in my side. "Should" have worked, but:

~~~
Parameters:
cmake
C:\WITH SPACES IT IS\Library\HelloWorld-CMake
-G
Ninja
-B
C:\WITH SPACES IT IS\Library\HelloWorld-CMake/build
-DCMAKE_BUILD_TYPE=RelWithDebInfo
-DBUILD_TESTING=YES
-D
CMAKE_Swift_FLAGS=-sdk "C:\WITH SPACES IT IS\Library\Developer\Platforms\Windows.platform\Developer\SDKs\Windows.sdk" -I "C:\WITH SPACES IT IS\Library\Developer\Platforms\Windows.platform\Developer\SDKs\Windows.sdk/usr/lib/swift" -L "C:\WITH SPACES IT IS\Library\Developer\Platforms\Windows.platform\Developer\SDKs\Windows.sdk/usr/lib/swift/windows"
~~

Error:

~~~
CMake Error: The source directory "C:/WITH SPACES IT IS/Library/IS/Library/Developer/Platforms/Windows.platform/Developer/SDKs/Windows.sdk/usr/lib/swift/windows" does not exist.
~~~

Hmm. This path is complete nonsense.

stefanspringer1: ...Using single quotes instead helps a little (CMAKE_Swift_FLAGS=-sdk 'C:\...), but then the compiler test fails:

~~~
 The Swift compiler

    "C:/WITH SPACES IT IS/Library/Developer/Toolchains/unknown-Asserts-development.xctoolchain/usr/bin/swiftc.exe"

  is not able to compile a simple test program.

...

<unknown>:0: error: no such file or directory: 'SPACES'
    <unknown>:0: error: no such file or directory: 'IT'
    <unknown>:0: error: no such file or directory: 'IS\Library\Developer\Platforms\Windows.platform\Developer\SDKs\Windows.sdk''
    <unknown>:0: error: no such file or directory: 'SPACES'
    <unknown>:0: error: no such file or directory: 'IT'
~~~

stefanspringer1: ...So changed text again to "...they should not contain spaces". I think we should publish it as it is, the space thing is a separate issue – it should be resolved, but the current changes to the Windows.md should not be hold back.

compnerd: I think that you need to use the correct escape character: ^ not \ like Unix.

compnerd:
What do you think of splitting this up?  Let’s split out the bits not related to building and get those merged while we work on improving the build instructions.

docs/Windows.md
compnerd: Okay, doing that as a follow up is fine.

compnerd:
Okay, since you think its better to merge this in this state, I don't think that its really that terrible to do that, lets get this merged and we can iterate on improving it further.

PR #184, opened 2020-03-25
caches: add `lldb-python-scripts`

compnerd:
This adds a missing component to the packaged tarball for the toolchain.


compnerd:
CC: @mikkeyboi - this at least builds it - need to get that into the installers still

PR #188, opened 2020-03-31
Use correct property name in PATH setter

wjk:
Without this change, the value added to the PATH will not be property rooted (i.e. it is missing the "C:" part at the beginning). This prevents the computer from finding either `swiftc.exe` or the Swift runtime binaries, making the files installed by these packages rather useless. Manually fixing the PATH using System Properties will work, but I wouldn’t recommend it as then the uninstaller will be unable to correctly remove the problematic entries (as the exact values it searches for will no longer exist).

I ran into this problem after installing the 5.2 compile release from this repo, and trying to compile the HelloWorld example project in swift-build-examples. I am not 100% positive that this is the correct fix, because I don’t (yet) know how to build the installers from scratch myself, but I’m pretty certain that these changes are correct. Thanks!

### DIFF
--- a/wix/windows-runtime.wxs
+++ b/wix/windows-runtime.wxs
@@ -94,7 +94,7 @@
 
     <DirectoryRef Id="TARGETDIR">
       <Component Id="ENV_VARS" Guid="879f2029-84e0-4c64-9595-c11c499ff340">
-        <Environment Id="PATH" Action="set" Name="PATH" Part="last" Permanent="no" System="yes" Value="[INSTALLDIR]\Library\Swift-development\bin" />
+        <Environment Id="PATH" Action="set" Name="PATH" Part="last" Permanent="no" System="yes" Value="[TARGETDIR]\Library\Swift-development\bin" />
       </Component>
     </DirectoryRef>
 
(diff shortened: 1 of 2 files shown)


compnerd:
Thanks @wjk!

compnerd:
Interesting that no one else has reported this and the installer works fine for me?  In any case, I've merged it for now (as I didn't realize that you hadn't tested it).

Building the installer is pretty much just extracting the large zip file of the toolchain (toolchain-windows-x64) and you can use `msbuild` with the parameters for the paths.  https://github.com/compnerd/swift-build/blob/master/wix/windows-toolchain.wixproj actually is a MSBuild file, you should be able to specify `TOOLCHAIN_ROOT` to give the root.  https://github.com/compnerd/swift-build/blob/master/.ci/templates/toolchain-msi.yml would be a good reference for the MSI build for the toolchain.

I think something along the lines of:
```
msbuild windows-toolchain.wixproj -p:OutputPath=S:\BinaryCache\msi -p:IntermediateOutputPath=S:\BinaryCache\msi\toolchain -p:TOOLCHAIN_ROOT=%USERPROFILE%\Downloads\windows-toolchain-x64\Library\Developer\Toolchains\unknown-Asserts-development.xctoolchain
```

compnerd:
Doing an experimental build at https://dev.azure.com/compnerd/swift-build/_build/results?buildId=29223&view=results ... this is for master, but should validate if the changes are good.

compnerd:
@wjk sorry, I had to revert the change.  The issue is that `TARGETDIR` is initialized the disk with the most free space, which is not the same as the install location.  This results in the paths actually being incorrect.

PR #190, opened 2020-03-31
install: prefix paths with the volume

compnerd:
Adjust the path to include the volume name as pointed out by @wjk.


compnerd:
CC: @wjk
Lets see if this works

compnerd:
Locally verified with this version that it adds the volume label to the path as well.  Going to do a build on CI as well.

PR #192, opened 2020-04-01
Add merge modules for Swift runtime and ICU

### LINKED ISSUE #186: Provide Merge Module for Swift runtime
I noticed while looking over the Releases tab that both an EXE bootstrapper and raw MSI files are present. I was wondering if you could also provide a [merge module](https://docs.microsoft.com/en-us/windows/win32/msi/merge-modules) for the Swift runtime. The benefit of this is that merge modules can be embedded within other MSI files, as opposed to top-level MSI files, which can only be bundled with other MSIs using an EXE bootstrapper. I prefer to ship raw MSI files to my users instead of EXE bootstrappers because (a) the built-in MSI install-time UI is much more usable than the one provided by default by the WiX bootstrapper subsystem, and (b) raw MSIs are better supported in enterprise deployment scenarios as compared to installer EXEs. Merge modules are reference-counted by the system, and will not be uninstalled until the last top-level product that embeds them is removed.

You can build a merge module using [the WiX `Module` element](https://wixtoolset.org/documentation/manual/v3/xsd/wix/module.html). I would then suggest editing the runtime MSI file to embed the runtime merge module, and contain no other files.

If you need a hand with this, I can create a PR; I have experience creating and embedding MSMs from one of my non-Swift projects. I just wanted to the idea by you before I started work. Thanks!

wjk:
Does what the title says. Tested on my dev box and confirmed working. You’ll want to add the .MSM files to the next release you perform; they cannot be extracted from the MSI they are merged into. I have not integrated this into the CI build; you will have to do this after the PR is merged. Enjoy!

Fixes #186.


compnerd:
Hmm, Im not sure I fully understand this.  In particular, what does it mean for them to be unable to be extracted from the MSI?  Why no MSM for the toolchain?  I think that we should have a single MSI installer with MSMs for each of the components that have the MSIs.

wjk:
MSIs and MSMs, though very similar, are distinct file formats that cannot be interchanged. WiX (and Windows Installer underneath it) simply will not allow you to merge one complete MSI into another; you must use an MSM file. [Microsoft docs explaining this are here](https://docs.microsoft.com/en-us/windows/win32/msi/merge-module-database).

Furthermore, I know of no tool that can recreate an MSM file given only the MSI file it was merged into. Even though the contents of MSM files are ref-counted separately from their host MSI (so that an MSM containing components shared between different applications is only removed when the last application referencing that MSM is uninstalled), the contents of the MSM tables are merged directly into the tables of the host MSI file. To a tool such as [my MSI Viewer app](https://github.com/wjk/MSIExtractApp), the contents of the MSM and the host MSI are listed side-by-side, as one physical group. I hope this explains this issue.

I will convert the other MSIs to MSMs, and create one large MSI combining them, sometime in the next few days. Thanks!

wjk:
And done! There is now one, and only one MSI file generated, and it consists of nothing but MSM references. I have tested the resulting (300+MB) MSI file on my local dev box, and it installs correctly.

I also took the opportunity (in 3af893c28e3cd8fe0986d0f7e80a7afa944ebf7e) to parameterize the upgrade code for the TensorFlow and non-TensorFlow versions of the MSI. With this change applied, I can install the regular Swift MSI, then the TensorFlow MSI, and then uninstall the regular MSI, and all of the components installed by both will still be present! Without this change, the user must completely uninstall the regular MSI before they can install the TensorFlow MSI, and vice-versa. Enjoy!

compnerd:
Awesome, thanks @wjk! I'll look into these changes once the CI has been repaired.  Seems that there were a series of regressions recently, and there is just a lot of repair work that needs to be done.

wix/windows-runtime-msm.wxs
```diff
@@ -0,0 +1,84 @@
...
+             Description="Windows Swift Runtime for x86_64"
+             InstallScope="perMachine"
+             Manufacturer="dt.compnerd.org" />
+
+    <!-- Directory Structure -->
+    <Directory Id="TARGETDIR" Name="SourceDir">
+      <Directory Id="WINDOWSVOLUME">
+        <Directory Id="LIBRARY" Name="Library">
+          <Directory Id="SWIFT" Name="Swift-development">
+            <Directory Id="BIN" Name="bin">
```
compnerd: Might be a good time to inject `USR` into the path for uniformity.

wix/windows-runtime-msm.wxs
```diff
@@ -0,0 +1,84 @@
...
+        <File Id="SWIFTTENSORFLOW_PDB" Source="$(var.SDK_ROOT)\usr\bin\swiftTensorFlow.pdb" Checksum="yes" DiskId="2" />
+        <File Id="SWIFT_DIFFERENTIATION_PDB" Source="$(var.SDK_ROOT)\usr\bin\swift_Differentiation.pdb" Checksum="yes" DiskId="2" />
+        <?endif?>
+      </Component>
+      <?endif ?>
+    </DirectoryRef>
+
+    <DirectoryRef Id="TARGETDIR">
+      <Component Id="ENV_VARS" Guid="879f2029-84e0-4c64-9595-c11c499ff340">
+        <Environment Id="PATH" Action="set" Name="PATH" Part="last" Permanent="no" System="yes" Value="[WindowsVolume]Library\Swift-development\bin" />
```
compnerd: This would need to be updated accordingly.

wix/windows-icu-msm.wxs
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Wix xmlns="http://schemas.microsoft.com/wix/2006/wi">
+  <Module Id="dt.compnerd.org.SwiftRuntime.ICU" Language="1033" Version="64.1">
```
compnerd: Is there a reason to put this into the `SwiftRuntime` namespace rather than having a top-level namespace?  Id be fine with `SwiftRuntimeICU` to differentiate it from a standard ICU build.

wix/windows-icu.wxs
```diff
@@ -36,12 +32,8 @@
...
 
     <!-- Components -->
-    <DirectoryRef Id="ICU_USR_BIN">
-      <Component Id="ICU_RUNTIME" Guid="af2bccab-70c7-41cd-9ec6-a265f79c0d63">
-        <File Id="ICUDT64_DLL" Source="$(var.ICU_ROOT)\Library\icu-64\usr\bin\icudt64.dll" Checksum="yes" />
-        <File Id="ICUIN64_DLL" Source="$(var.ICU_ROOT)\Library\icu-64\usr\bin\icuin64.dll" Checksum="yes" />
-        <File Id="ICUUC64_DLL" Source="$(var.ICU_ROOT)\Library\icu-64\usr\bin\icuuc64.dll" Checksum="yes" />
-      </Component>
+    <DirectoryRef Id="ICU">
+      <Merge Id="icu-msm" Language="1033" DiskId="1" SourceFile="$(var.windows-icu-msm.TargetPath)" />
```
compnerd: Why not handle this like the runtime and merge into `Library` rather than merge into `Library\icu-64`?

wix/windows-runtime.wxs
```diff
@@ -1,44 +1,84 @@
...
-    <?ifdef INCLUDE_DEBUG_INFO ?>
-    <Media Id="2" Cabinet="runtime_PDBs.cab" EmbedCab="yes" />
-    <?endif?>
-
     <!-- Directory Structure -->
     <Directory Id="TARGETDIR" Name="SourceDir">
       <Directory Id="WINDOWSVOLUME">
         <Directory Id="LIBRARY" Name="Library">
-          <Merge Id="runtime-msm" DiskId="1" Language="1033" SourceFile="$(var.windows-runtime-msm.TargetPath)" />
+          <Directory Id="SWIFT" Name="Swift-development">
```
compnerd: Bleh, this really should get versioned (`Swift-5.2`, `Swift-5.3` and `Swift-development`)

wix/windows-installer.wixproj
```diff
@@ -0,0 +1,49 @@
...
+
+  <PropertyGroup>
+    <WixTargetsPath Condition=" '$(WixTargetsPath)' == '' ">$(MSBuildExtensionsPath)\Microsoft\WiX\v3.x\Wix.targets</WixTargetsPath>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <ProjectGuid>6c7a4772-77e1-4d3e-a7e8-ebf01a8965ef</ProjectGuid>
+    <ProductVersion>0.0.0</ProductVersion>
+    <OutputName Condition="'$(TENSORFLOW)' == 'true'">SwiftForTensorflow</OutputName>
+    <OutputName Condition="'$(TENSORFLOW)' != 'true'">Swift</OutputName>
```
compnerd, line 16: This is horrid, thank you XML.  Its a clever trick though :)

compnerd:
@wjk could you please rebase this PR so that I can run it through CI in the mean time?

wjk:
@compnerd Just circling back around to this PR since I have a moment. This really needs to be closed, as I now realize that it was based on an incorrect assumption of the workings of WiX and MSM.

The problem is, the component GUIDs (for each individual file) **must** remain the same, always, for that specific file across all builds of all versions of the same MSM file. Otherwise, Windows would think that two copies of, say, `

PR #193, opened 2020-04-01
Add WiX signing support

### LINKED ISSUE #151: Enable Signing for MSIs
We should sign the MSIs that we are generating for the toolchain.

wjk:
Does what the title says. Note that this is necessarily incomplete. For it to be production-ready, you’ll have to replace the `CertificateThumbprint`, `SignDescription`, and `SignInformationUrl` properties in CodeSigning.props with appropriate values. Then pass `/p:SigningEnabled=true` when building one of the WiX projects to sign all build artifacts as they are created.

Don’t forget to add the imports to `windows-runtime-msm.wixproj` and `windows-icu-msm.wixproj` once #192 is merged, or the merge modules will not be signed.

Fixes #151.

---
This change is [Reviewable](https://reviewable.io/reviews/compnerd/swift-build/193)

### DIFF
--- a/wix/targets/CodeSigning.Tasks.targets
+++ b/wix/targets/CodeSigning.Tasks.targets
@@ -1,0 +1,201 @@
+<Project>
+  <UsingTask TaskName="CheckIfCertificateIsInstalled" TaskFactory="RoslynCodeTaskFactory" AssemblyFile="$(MSBuildToolsPath)\Microsoft.Build.Tasks.Core.dll">
+    <ParameterGroup>
+      <CertificateThumbprint ParameterType="System.String" Required="true" />
+      <CertificateExists ParameterType="System.Boolean" Output="true" />
+      <CertificateStoreLocation ParameterType="System.String" />
+      <CertificateStoreName ParameterType="System.String" />
+    </ParameterGroup>
+
+    <Task>
+      <Using Namespace="System.Security.Cryptography.X509Certificates" />
+      <Code Type="Fragment" Language="cs">
+        <![CDATA[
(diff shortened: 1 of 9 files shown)


compnerd:
Is there a way to simply give the installer a signing certificate rather than having to embed the thumbprint and everything?  The Azure setup has the secrets to sign the binary (its a sign tool compatible PKCS).

wjk:
The thumbprint is not actually embedded; it is given to signtool.exe so it can locate the correct certificate (as several code-signing certificates with private keys can conceivably exist). Please see [the documentation for signtool.exe](https://docs.microsoft.com/en-us/windows/win32/seccrypto/signtool). If you scroll down to the description of the `/sha1` flag,

compnerd:
Hmm, is there a reason to build the custom signature handling rather than using `insignia` from WiX?  The PKCS is the same thing that you can generate from the windows tools, its just that I have that stored securely on Azure so that I can sign the macOS build and would like to reuse the certificate for Windows as well.

wjk:
Again, this code is what works for me. I don't use Azure code signing. However, since we override the `Sign*` targets like I have done in this PR, running insignia is not required, as the relevant bits and pieces will be signed _in situ_ as the final product is assembled (before the CAB is added to the MSI, before the MSI is embedded into the bootstrapper, before the bootstrapper is packed into the stub EXE, and so on).

compnerd:
Unfortunately, I don’t think that can work for the toolchain, Microsoft documents that if you have large binaries (like this case), you should use a manifest and sign that.  I’ll try to find the documentation for that on MSDN again.

PR #204, opened 2020-04-13
DNM: trace foundation configuration

compnerd:


compnerd:
Not digging into this currently, going to close this off for now.

PR #205, opened 2020-04-14
android: force configuration

compnerd:
For some reason the `CMAKE_HAVE_LIBC_PTHREAD` is failing on the builder.
Force the setting to repair the build.


compnerd:
Seems to to be better than before, will have to address the x86 failure separately.

PR #230, opened 2020-05-16
Typo

save-buffer:


compnerd:
Thanks @save-buffer!  I still type gud :-p

PR #233, opened 2020-05-19
CI: attempt to update ICU to to 67.1

compnerd:
Update the ICU build to 67.1 experimentally.  If this actually repairs
the ICU build, it may be valuable to jump to it.


compnerd:
There are a number of issues on the CI currently, I think that the complexity of trying to get this right rather than fixing forward is higher.  At this it will make it possible to bring up individual targets without re-building everything. 

This is probably going to impact the Swift 5.2 builds which need ICU-64

PR #240, opened 2020-05-28
CI: externalize ICU configuration

compnerd:
Hoist the ICU version selection to metadata on the branch.  This allows
us to use different versions of ICU for different builds.  This is
required in order to bump to ICU 67.


compnerd:
The TensorFlow builds are failing even without this change, and the Ubuntu one is failing on the devtools builds due to Yams, going to merge this.

PR #253, opened 2020-07-09
Update Windows docs to use the system drive to avoid compile failures

alex-taffe:
Fixes https://github.com/compnerd/swift-build-examples/issues/20. As pointed out by @compnerd, the build command may fail if a drive name is not provided for the compiler. `%SystemDrive%` was chosen to give new users the largest chance of success when compiling.


compnerd:
Seems reasonable to use `%SystemDrive%`.  I wish that there was a better way :-(

compnerd:
Thanks for the change!

PR #276, opened 2020-09-05
CI: Use Python 3 instead of 2

stevapple:
This change is [Reviewable](https://reviewable.io/reviews/compnerd/swift-build/276)

### DIFF
--- a/.ci/templates/windows-devtools.yml
+++ b/.ci/templates/windows-devtools.yml
@@ -1,3 +1,7 @@
+parameters:
+  - name: PythonVersion
+    type: string
+    default: '3.7.x'
 
 jobs:
   - job: ${{ parameters.host }}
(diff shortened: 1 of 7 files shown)


compnerd:
Thanks for the PR, however, I think that this would break.  This doesn't account for the older builds which require python 2.  Could you please wire it up such that python 2 is used for 5.2/5.3 and python 3 is used for newer versions?

compnerd (changes requested):
Thanks for the PR, however, I think that this would break. This doesn't account for the older builds which require python 2. Could you please wire it up such that python 2 is used for 5.2/5.3 and python 3 is used for newer versions?

compnerd (approved):
Awesome, I really like this.  Lets get a couple of builds tested and get this merged!

.ci/templates/windows-devtools.yml
```diff
@@ -1,3 +1,7 @@
+parameters:
```
compnerd, line 1: I think that this part of the change is causing the issue.  When you list parameters, you need to list all of them :-(

compnerd:
The packaging steps are expected to fail on the test; as long as the builds are good, I should be able to merge this.  Thanks for the changes here @stevapple!

PR #287, opened 2020-09-10
CI: add a couple of plists

compnerd:
These will communicate information to swift-package-manager.

---
This change is [Reviewable](https://reviewable.io/reviews/compnerd/swift-build/287)


lxbndr:
Just curious, what these plist do?

compnerd:
The plist itself is just a storage mechanism.  It contains information relevant to the SDK/platform support (i.e. the XCTest version and the MSVC runtime in use).  This information is useful to external consumers (i.e. s-p-m).  They are intended to serve as a bridge to s-p-m to enable `swift build` to be used without any flags.

PR #292, opened 2020-09-17
build: Fix parameter for "check-swift"

stevapple:
Finally fixed (hopefully) `check-swift` failure by fixing a typo in build parameters.

---
This change is [Reviewable](https://reviewable.io/reviews/compnerd/swift-build/292)

### DIFF
--- a/.ci/templates/toolchain.yml
+++ b/.ci/templates/toolchain.yml
@@ -337,7 +337,7 @@ jobs:
             -D CLANG_TABLEGEN=$(CLANG_TABLEGEN)
             -D LLDB_TABLEGEN=$(LLDB_TABLEGEN)
             -D SWIFT_PATH_TO_LIBDISPATCH_SOURCE=$(Build.SourcesDirectory)/swift-corelibs-libdispatch
-            -D SWIFT_NATIVE_SWIFT_TOOLS_PATH=$(Build.BinariesDirectory)/llvm-tools/bin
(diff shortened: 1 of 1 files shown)


compnerd:
This wasn't a typo.  In order to build the Windows ARM64 target, I need to pass the swift tools for the host tools that are built to build the diagnostics.  The build needs to fallback to the just built tools if it can instead.

compnerd:
https://github.com/apple/swift/pull/33915 should fix it though (however, it now conflicts due to other changes)

stevapple:
But the passed dir is `llvm-tools/bin`, and flag is `SWIFT_NATIVE_SWIFT_TOOLS_PATH`? This directory’s doesn’t contain Swift tools like `swiftc`, so `check-swift` won’t work properly.

compnerd:
Yes, the path is `llvm-tools/bin` and that *is* the swift tools

stevapple:
> [apple/swift#33915](https://github.com/apple/swift/pull/33915) should fix it though (however, it now conflicts due to other changes)

I see. I think that shall fix it. Waiting for it to be merged.

stevapple:
Is there any way to reach to you other than mentioning on GitHub (if possible, by DM or mail)? I may need some help in dealing with Windows problems since I’m not an expert here. @compnerd

compnerd:
Email tends to work well (its pretty well littered and you should be able to find it in the git history as well - lets not make it easier for bots to scrape it than it already is)

PR #295, opened 2020-09-23
build: further tweak layout

compnerd:
Adjust the layout to allow the default resource path to be sufficient to
find the modules.  This reduces the flags necessary for Windows to
`-sdk %SDKROOT%`.

---
This change is [Reviewable](https://reviewable.io/reviews/compnerd/swift-build/295)


compnerd:
This doesn't break 5.3, but isn't sufficient enough either.  I'll do the rest for 5.3.1 in a follow up.

PR #298, opened 2020-09-24
WIX: autogenerate component guids at compile time

compnerd:
Adjust the GUID generation to be compile time computation rather than
immediate.  This should allow installing over in future versions.

---
This change is [Reviewable](https://reviewable.io/reviews/compnerd/swift-build/298)


compnerd:
https://github.com/apple/swift-installer-scripts/pull/207/commits/59874fbd89770ae1f6a2374a5a1a638f02b9d456 should accomplish this.

PR #309, opened 2020-10-13
WIX: allow user to customize install location

compnerd:
This change is [Reviewable](https://reviewable.io/reviews/compnerd/swift-build/309)

### DIFF
--- a/wix/windows-bundle.wxs
+++ b/wix/windows-bundle.wxs
@@ -17,22 +17,39 @@
       <bal:WixStandardBootstrapperApplication
         LicenseUrl=""
         LogoFile="$(var.Logo)"
-        SuppressOptionsUI="yes"
+        SuppressOptionsUI="no"
         SuppressRepair="yes" />
     </BootstrapperApplicationRef>
+
+    <Variable
+      Name="INSTALLDIR"
+      Type="string"
+      Value="[WindowsVolume]"
+      bal:Overridable="yes" />
 
     <Chain>
       <MsiPackage SourceFile="$(var.MSI_LOCATION)\toolchain.msi" Compressed="yes">
         <MsiProperty Name="INSTALL_DEBUGINFO" Value="[INSTALL_DEBUGINFO]" />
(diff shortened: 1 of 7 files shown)


wix/windows-bundle.wxs
```diff
@@ -17,22 +17,39 @@
...
         LogoFile="$(var.Logo)"
-        SuppressOptionsUI="yes"
+        SuppressOptionsUI="no"
         SuppressRepair="yes" />
     </BootstrapperApplicationRef>
 
+    <Variable
+      Name="INSTALLDIR"
+      Type="string"
+      Value="[WindowsVolume]"
```
wjk, line 27:
```suggestion
      Value="[WindowsVolume]\Library"
```

wjk (changes requested):
I recommend that you change the default `INSTALLDIR` value from `[WindowsVolume]` to `[WindowsVolume]\Library`. If users choose to install the Swift runtime into a different folder, chances are it will be a newly created, otherwise-empty folder. This way, there won't be an otherwise-useless Library folder inside that folder.

Note that if you choose to do this, you will have to get rid of all the `<Directory Id="LIBRARY" Name="Library">` tags in the MSI projects to match the change in the bundle project. (Everything inside those tags can be kept as-is. Just delete the lines containing the opening and closing tags for the Library directory element, then auto-reindent.)

compnerd:
Hmm, @wjk I think you have a good point.  However, I'm not convinced that should be part of this change.  Instead, we should make that change even if we don't actually manage to get this particular change merged because that actually improves the installer.

wjk:
So you're saying that we should change the INSTALLDIR in a later PR? Your message is kinda confusing.

compnerd:
I'm suggesting that we should do it in a different PR, but do that before this change.  I think that the change you are proposing will actually improve the installer for the extract only case and help with this PR.  Sorry about the message being confusing.

compnerd:
I'm going to close this for now.  This is pretty out of date, and more importantly, the installer content has moved to https://github.com/apple/swift-installer-scripts.

PR #311, opened 2020-10-19
gitignore installer files

ataffeqeexo:
Add `installer.exe` and `installer.exe.zip` to the gitignore list to prevent them from being accidentally committed to the repo

---
This change is [Reviewable](https://reviewable.io/reviews/compnerd/swift-build/311)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,4 @@
 # *~
 
 .*.sw[nop]
-utilities/installer.exe*
+installer.exe*
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,4 @@
 # *~
 
 .*.sw[nop]
-
+utilities/installer.exe*


compnerd:
I'm not convinced from the diff that this is useful.  The downloaded file can end up anywhere.  If you tend to download the installer into the current directory, you could always use `core.excludesfile` globally to augment the ignore list.  Is there something that I'm missing?

ataffeqeexo:
@compnerd I guess making it directory agnostic makes more sense. I could add it to the `core.excludesfile` but seems like it just makes more sense to have it in the repo so that people don't accidentally commit during a PR or something similar

alex-taffe:
True, I guess I could go either way. I was just testing the download script with the last `--quiet` PR and accidentally committed the installer with `git add .`. It seems like it couldn’t hurt to add it, but I also understand not wanting to unnecessarily junk up the `.gitignore` file

PR #328, opened 2020-12-09
add Dockerfile for windows based development image

vprimachenko:
hi compnerd, first and foremost huge thank you for all the work you did to bring swift to windows!

i noticed that the official download page of swift lacks a docker image for windows development and CI/CD

this commits adds a `windows/servercore:1809` based development image containing the buildtools and a swift toolchain from `//swift/build`

sadly to fulfill dockers buildcontext requirements i had to duplicate the `swift-build.py` file. it should be possible to clean this up later

### DIFF
--- a/docker/Windows-1809/Dockerfile
+++ b/docker/Windows-1809/Dockerfile
@@ -1,0 +1,37 @@
+# escape=`
+
+FROM mcr.microsoft.com/windows/servercore:1809-amd64
+
+# Install windows buildtools.
+RUN powershell mkdir .\TEMP\; `
+    Invoke-WebRequest -Uri https://aka.ms/vs/16/release/vs_buildtools.exe -OutFile .\TEMP\vs_buildtools.exe; `
+    Start-Process .\TEMP\vs_buildtools.exe  -Wait -ArgumentList '--quiet --wait --norestart --nocache `
+    --installPath C:\BuildTools `
+    --add Microsoft.VisualStudio.Component.VC.CMake.Project `
+    --add Microsoft.VisualStudio.Component.VC.Tools.x86.x64 `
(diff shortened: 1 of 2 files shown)


compnerd:
Thanks!  I like the idea of the Windows docker image, however, I don't think that I can have these be uploaded to docker hub correct?

I think that it would be better to just create a symbolic link for `swift-build.py`.  IIRC, the download using powershell is pretty slow and requires additional arguments to get reasonable donwload speeds.  Did you not run into that?

vprimachenko:
> Thanks! I like the idea of the Windows docker image, however, I don't think that I can have these be uploaded to docker hub correct?

as long as you have a docker hub account and are member of the [swift namespace](https://hub.docker.com/_/swift). alternatively you can offer them from your own azure registry. as to the license to use windows base image it should be ok, [docker themselves](https://hub.docker.com/_/hello-world/) offer a [demo image based on windows](https://github.com/docker-library/hello-world/blob/9c93e37114a7fe99b5fc0d776e0b8dff99cbbb75/amd64/hello-world/nanoserver-1809/Dockerfile) 

> I think that it would be better to just create a symbolic link for `swift-build.py`. 

will look into that. did not know that git can handle a symlink

> IIRC, the download using powershell is pretty slow and requires additional arguments to get reasonable donwload speeds. Did you not run into that?

the only things downloaded via powershell are the installer for buildtools (couple MB) and the one for azure. on local tests and using gitlab ci there were no issues, only thing that was slow - was the download of `installer.exe` from azure

compnerd:
This was never addressed, and at this point, I'm thinking that we no longer want to add this here.  If you are interested in this still (I am, and have made some progress towards this too), you should contribute the improvements to apple/swift-docker.

Thanks!

PR #331, opened 2020-12-14
Point tensorflow repos to `main`

texasmichelle:
This change is [Reviewable](https://reviewable.io/reviews/compnerd/swift-build/331)


compnerd (changes requested):
The default revision change requires a larger change; I can push that part of the change and then you could rebase on top of that.

compnerd:
The S4TF project has been archived 😭

PR #350, opened 2021-02-19
WIX: enable Developer Mode with SwiftPM

stevapple:
SwiftPM requires Developer Mode to function properly. Instead of asking users to enable it manually, we can enable it through the installer.

---
This change is [Reviewable](https://reviewable.io/reviews/compnerd/swift-build/350)

### DIFF
--- a/wix/windows-devtools.wxs
+++ b/wix/windows-devtools.wxs
@@ -149,7 +149,7 @@
 
         <?ifdef INCLUDE_DEBUG_INFO ?>
         <Component Id="SWIFT_CRYPTO_DEBUGINFO" Guid="4cbda17f-a85b-4620-bdd4-18b03ba07499">
-          <File Id="CRYPTO_PDB" Source="$(var.DEVTOOLS_ROOT)\usr\bin\Crypto.pdb" Chekcsum="yes" />
+          <File Id="CRYPTO_PDB" Source="$(var.DEVTOOLS_ROOT)\usr\bin\Crypto.pdb" Checksum="yes" />
         </Component>
         <?endif?>
       <?endif?>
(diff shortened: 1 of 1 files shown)


stevapple:
A brief explanation on what SwiftPM uses this for: SwiftPM needs Developer Mode to create symbolic links with the `SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE` flag (see [CreateSymbolicLinkW](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-createsymboliclinkw)), or it will throw an error on every successful build.

compnerd (changes requested):
The fixes beyond the developer mode are good, can you split that out into a separate change?  The problem with the developer mode that I have is that it isn't entirely required.  I do not have developer mode enabled on my machine but am using swift package manager just fine.  I am uncomfortable with such a change.  I would rather that this be an explicit opt-in via a checkbox option in an advanced setting.

stevapple:
It’s really confusing though. May you check the key `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\AppModelUnlock\AllowDevelopmentWithoutDevLicense` on your machine? If the value is 1, there should be other application that already enables it.

compnerd:
```
>reg query HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\AppModelUnlock\AllowDevelopmentWithoutDevLicense
ERROR: The system was unable to find the specified registry key or value.
```

It is not confusing at all, it is well documented at https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-createsymboliclinkw

stevapple:
According to the document and the implementation of `swift-corelibs-foundation` and `swift-package-manager`, there should always be a `SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE` flag passed to `CreateSymbolicLinkW`. If Developer Mode isn’t enabled, there should be error thrown. Did I miss any point?

compnerd:
`SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE` does not mean that you can create the symbolic link without the privilege.  It means that the process does not need to be running in an elevated context.  You still need `SE_ALLOW_CREATE_SYMBOLIC_LINK_PRIVILEGE`.  Enabling Developer Mode will reduce the security on the system allowing untrusted code execution amongst other things.  One of the things it enables is effectively every process receiving that security permission, thus allowing the creation of symbolic links.

I've generally been very intentional about the Windows port.

compnerd:
Further verified: https://forums.swift.org/t/stochastic-errors-resolving-dependencies-and-or-building-with-feb-18-2021-snapshot

The developer mode is most definitely not the correct thing, and I don't think that is something that the installer should be doing by default.

stevapple:
In fact, I’ve already tried to alter user permission through WiX, and that seems impossible. There’s not even a way to grant that through the shell AFAIK.
Using Developer Mode is one of the solutions, and maybe the only one possible for an installer.

stevapple:
Another point is that there’s no `secpol.msc` on Windows 10 Home. Developer Mode  (by registry key), meanwhile, is available on all versions.

compnerd:
Application through a GPO should be possible to automate.  You should be able to create a C# Custom Action to apply the GPO.

PR #355, opened 2021-04-14
CI: use older cmake (3.19.7)

lxbndr:
Current version (3.20) is known to have some issues with Swift pipelines.
3.19.7 is a last version we had before builds gone red.

I suggest to switch back to 3.19 until we have a new CMake release.

---
This change is [Reviewable](https://reviewable.io/reviews/compnerd/swift-build/355)


compnerd (approved):
Thanks!

PR #356, opened 2021-04-15
CI: install cmake on Windows agent only

lxbndr:
I absolutely forgot about non-Windows agents when did #355. We should skip that step on other systems, like macOS, etc.

---
This change is [Reviewable](https://reviewable.io/reviews/compnerd/swift-build/356)


compnerd (approved):
This works, but I don't think that it is sufficient.  This needs to be restricted to hosted agents only.

lxbndr:
Ah, I see. Will do as follow up.

PR #357, opened 2021-04-15
CI: install cmake only on hosted agents

lxbndr:
I am not sure if this is correct way to check for agent type. Can't find any specific variable for such agents, unfortunately. Looks like the only way is to check agent name. Guessing from recent jobs, there are "Hosted Agent" and "Azure Pipelines N" agents.

---
This change is [Reviewable](https://reviewable.io/reviews/compnerd/swift-build/357)


compnerd (approved):
Thanks!

PR #411, opened 2021-12-01
Fix toolchain modules restructure step in build.cmd

lxbndr:
According to the actual directory tree and installer rules, we have to look for internal modules in `%ToolchainInstallRoot%\usr\lib\swift\`


compnerd (approved):
Thanks!

PR #414, opened 2021-12-22
Foundation build workaround for CMake 3.22

lxbndr:
CMake 3.22 fails to generate Foundation build files:
```
CMake Error in CoreFoundation/CMakeLists.txt:
  MSVC_RUNTIME_LIBRARY value 'MultiThreadedDLL' not known for this ASM
  compiler.
```

I am still not sure _why_ it fails to detect the correct MSVC library flag. The [code related to this option](https://gitlab.kitware.com/cmake/cmake/-/blob/master/Source/cmLocalGenerator.cxx#L1968) seems to be the same as for previous versions. The configuration step has the actual value of `CMAKE_ASM_COMPILE_OPTIONS_MSVC_RUNTIME_LIBRARY_MultiThreadedDLL`  set to `-MD`. But I guess we could give CMake a hint as a workaround.

### DIFF
--- a/.github/workflows/swift-toolchain.yml
+++ b/.github/workflows/swift-toolchain.yml
@@ -741,6 +741,7 @@ jobs:
                 -D CMAKE_C_FLAGS="/GS- /Oy /Gw /Gy" `
                 -D CMAKE_CXX_COMPILER=${CLANG_CL} `
                 -D CMAKE_CXX_FLAGS="/GS- /Oy /Gw /Gy" `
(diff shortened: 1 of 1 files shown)


compnerd:
I suspect that this might be due to the compiler detection changes that have occurred in the 3.22 release.  Since I can't trigger the CI from the PR, and it seems fine and won't actually regress anything more than the current state, lets merge this and go from there.

compnerd (approved):
This is a bit weird of a change, but if it fixes things, that is certainly better.   It would be nice to avoid this entirely though.

PR #415, opened 2021-12-23
NFC: Fix typo in toolchain config command

lxbndr:
Made this typo in #390, and since then it is an eyesore to me. Can't stand it anymore😅


compnerd (approved):
Thanks!

PR #423, opened 2022-02-17
build llvm-config as part of build tools job

lxbndr:
This is one of the steps required to eventually enable runtimes in arm64 builds. `builtins` and `compiler-rt` targets are depending on `llvm-config` tool. This triggers native llvm build If the tool is not provided externally. We could build `llvm-config` together with other native tools we already preparing in `build_tools` job.


compnerd (approved):
This looks good, thanks!

PR #430, opened 2022-05-15
VS2022, 5.7: use VC 14.31 toolset

lxbndr:
This is workaround for current miscompiling issue with latest MSVC toolset. Until the reason of the issue is known, we could use last known good toolset version.

Please note that I didn't actually test this solution with Pipelines, and I am not sure if it is possible to run build check against this PR. For GitHub Actions it worked well, so I believe this is the way we can fix builds temporarily, but I'd like to apologize in advance for any possible typos or yml syntax errors. I'll be here to fix them in follow-ups.

### DIFF
--- a/.ci/vs2022-swift-5.7.yml
+++ b/.ci/vs2022-swift-5.7.yml
@@ -258,6 +258,12 @@ stages:
           - checkout: apple/swift-experimental-string-processing
             fetchDepth: 1
           - script: |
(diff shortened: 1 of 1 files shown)


compnerd (approved):
I think that this is a reasonable workaround.  We should apply this to main as well (vs2022.yml).  I don't think that we can test this without merging, so let's do that.

PR #431, opened 2022-05-16
VS2022, main and 5.6: use VC 14.31 toolset

lxbndr:
Same changes as in #430, but for main and 5.6 branches.


compnerd:
The test is going to fail due to the packaging failures (signing certificate is unavailable - for security reasons).  However, I expect that this is safe as a mitigation.  Thank you for the changes!

compnerd (approved):

PR #440, opened 2022-06-26
Fix 5.5, vs2022

lxbndr:
This change basically covers 4 areas:
- Adds `PYTHON_EXECUTABLE` and `SWIFT_WINDOWS_$(platform)_ICU_nnn` parameters for toolchain and SDK respectively.
- Installs and uses VC 14.31 toolset to workaround the issue with latest toolset version.
- Finds and configures external clang-cl to use with some parts of toolchain and sdk.
- Patches Swift project to use external clang-cl.

A bit more details about the last point. The Swift patch is similar to the one that introduced `SWIFT_CLANG_LOCATION` in main Swift branch. The patch also allows to specify `SWIFT_LIBDISPATCH_C/CXX_COMPILER` externally for libdispatch external project, because Dispatch doesn't compile with external clang.

### DIFF
--- a/.ci/vs2022-swift-5.5.yml
+++ b/.ci/vs2022-swift-5.5.yml
@@ -232,6 +232,8 @@ stages:
           - script: |
               git config --global --add core.autocrlf false
               git config --global --add core.symlinks true
+          - checkout: self
+            fetchDepth: 1
           - checkout: apple/llvm-project
             fetchDepth: 1
           - checkout: apple/swift-cmark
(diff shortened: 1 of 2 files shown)


lxbndr:
Oh, and packaging steps are using scripts from [apple/swift-installer-scripts](https://github.com/apple/swift-installer-scripts) and the 5.5 branch needs some fixes as well. Also, it looks like there is some issue with the secure file used in the ICU package.

lxbndr:
https://github.com/apple/swift-installer-scripts/pull/136

PR #441, opened 2022-08-11
vs2022: force python x64

lxbndr:
Latest builds are broken because agent uses python x86 for some reason. In the build logs I see this:
```
-- Found Python3: C:/hostedtoolcache/windows/Python/3.10.6/x86/python3.exe (found suitable version "3.10.6", minimum required is "3.6") found components: Interpreter 
```

Let's try to force the x64 version. I didn't actually try this with Swift pipeline, so there are chances it would not help. In theory, `UsePythonVersion` task would set up the requested Python version and add it to the PATH. And then CMake should just pick it up for toolchain config.

### DIFF
--- a/.ci/vs2022.yml
+++ b/.ci/vs2022.yml
@@ -237,7 +237,7 @@ stages:
             'x64':
               arch: amd64
               platform: x86_64
-              EXTRA_CMAKE_ARGS: 
+              EXTRA_CMAKE_ARGS: -D Python3_ROOT_DIR=$(PYTHON_ROOT) -D Python3_INCLUDE_DIR=$(PYTHON_ROOT)/include -D Python3_LIBRARY=$(PYTHON_ROOT)/libs/python310.lib
             'arm64':
               arch: arm64
               platform: arm64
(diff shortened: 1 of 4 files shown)


lxbndr:
Looks like CI doesn't check vs2022.yml pipeline. I added the same task to 5.5, 5.6 and 5.7.

lxbndr:
Well, now I think we have to define Python location explicitly. Toolchain config keeps using x86 disregarding x64 Python in the PATH.

compnerd (approved):
This seems reasonable.  It is unfortunate, but I suppose that repairing this is important.  I wonder if we should pin to the same version as VS, but we can do that subsequently.

.ci/vs2022.yml
```diff
@@ -283,6 +283,11 @@ stages:
               filename: $(VsDevCmd)
               arguments: -no_logo -arch=$(arch) -host_arch=amd64 -vcvars_ver=14.31
               modifyEnvironment: true
+          - task: UsePythonVersion@0
+            inputs:
+              versionSpec: '3.10' 
+              addToPath: true 
+              architecture: 'x64'
```
compnerd, line 290: Actually, I think that this needs to be `$(arch)`.  We also have the ARM64 toolchain.

compnerd (changes requested):

.ci/vs2022.yml
lxbndr: Oh, ARM64 would do fine. It is not even broken in main branch. We are using pythonarm64 package (with `EXTRA_CMAKE_ARGS` wiring it to the config), so ARM64 build is immune to any other Python installed on the build machine. It would be used as interpreter for codegen tasks, but lldb integration is using the correct arch.

I think I could make this task run only for x64 builds, because ARM64 doesn't need any python adjustment at the moment.

Unfortunately, `UsePythonVersion` task only supports x86 and x64. We could've implement a simple and common way to set up Python for all archs in such case.

PR #443, opened 2022-08-24
Adopt renamed swift-def tool converter

lxbndr:
Looks like `swift-def-to-yaml-converter` is now `swift-def-to-strings-converter`.


compnerd (approved):
Thanks!

PR #444, opened 2022-09-02
Clone swift-syntax Alongside Swift

CodaFi:
swift-syntax is going to own all of the infrastructure that
generates the syntax nodes, and the legacy libSyntax build
still depends on that infrastructure. To make this hang together,
we need to clone swift-syntax and point the compiler build into it.

This is being a little overly-aggressive by cloning swift-syntax
into strictly more places than is necessary. The idea being
that Swift itself will begin to require the Swift parser in more
parts of the build (cf Doug's changes to enable the "early swift syntax"
stage of the build).

### DIFF
--- a/.ci/vs2022.yml
+++ b/.ci/vs2022.yml
@@ -129,7 +129,11 @@ resources:
       name: unicode-org/icu
       ref: refs/heads/maint/maint-69
       type: github
-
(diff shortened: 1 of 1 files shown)


compnerd (approved):
Thanks!

PR #446, opened 2022-09-16
vs2022: use latest MSVC toolset

stevapple:
Tested locally with Visual Studio 17.3.4, it seems the toolset incompatibility problem has gone.

However, this release hasn't been rolled into the runner image yet, so we need to hold on to next release date to verify it on CI.

### DIFF
--- a/.ci/vs2022.yml
+++ b/.ci/vs2022.yml
@@ -267,16 +267,6 @@ stages:
             fetchDepth: 1
           - checkout: apple/swift-syntax
             fetchDepth: 1
-          - powershell: |
-              if ("$(arch)" -eq "amd64") {
-                $ArchComponent = ".x86.x64"
-                $AtlArchComponent = ""
-              } else {
-                $ArchComponent = ".ARM64"
-                $AtlArchComponent = ".ARM64"
-              }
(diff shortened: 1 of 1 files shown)


compnerd:
It appears that the current images on azure are running 17.3.32708.82 as per the documentation at actions/runner-images.  The next updates don't seem to indicate a newer toolset yet :-(.

stevapple:
So let’s see if the latest one is working? There could also be Swift-side fixes, and I freshly built the whole toolchain locally with 14.33 today and found everything just works.

stevapple:
So it’s reaching the time limit of 6 hours per job🤔 Had not expected that.

compnerd:
Yeah, the builders are very sensitive and we often run up against the time and space limitations.  That is part of why the windows builds are so strictly structured and controlled.

stevapple:
TLE again… It seems the building toolchain step is very likely to run out of time limit now (and it could get even longer). Is there a way to break it up?

compnerd:
The current split is about as fine grained as it can be without rebuilding things and risking things diverging in weird ways.  I would not be comfortable with splitting up the toolchain build any further.

stevapple:
I remembered that you said building Swift as part of LLVM will boost the performance by ~10%, so splitting them is, though technically possible, rather uncomfortable.

But the build time of Swift (actually LLVM I would say) is becoming more of a worry for CI… Shall I open an issue on apple/swift for this?

compnerd:
No, that is not the right way to handle this.  The build times on the Swift CI have already ballooned to 2x.  Increasing that further is not acceptable.  The way to handle this is to investigate why the builds are slower and addressing that.

compnerd:
The MSI build failures are expected for external contributors.

compnerd (approved):
Thanks!  Seems that the workaround is no longer needed, this does clean up the CI rules.

PR #448, opened 2022-10-20
Add SwiftSyntax to the SourceKit-LSP Build

CodaFi:
As of apple/sourcekit-lsp#651 sourcekit-lsp  now
depends on SwiftSyntax for its internal representation of documents.


compnerd:
The failures are related to codesigning, which is expected.  Thanks @CodaFi!

PR #450, opened 2022-12-13
[vs2022] build swift-compatibility-symbols tool

lxbndr:
Looks like we need one more tool now.


compnerd (approved):
Thanks!  Just made the same change for the GHA build.

PR #452, opened 2022-12-28
CI: switch Yams to static linking on Windows

compnerd:
```
dynamic:
  2,313,728 SwiftDriver.dll
    623,616 Yams.dll
static:
  2,712,064 SwiftDriver.dll
```

This saves 225,280 bytes in the actual release build.


compnerd:
cross-dependency: apple/swift-installer-scripts#155

PR #455, opened 2023-02-27
Refactor build.cmd to powershell

tristanlabelle:
Build.cmd contains a lot of repeated commands to build components for multiple architectures. This change refactors it to powershell to factor the repeated code as functions, improve debuggability with stack traces.


lxbndr:
I feel like I was using PS for build tasks countless times, and every single time I was running into some sort of inconvenience related to how powershell passes parameters to external commands. Quotation marks, quoting the quotation marks, spaces, line breaks, etc... One time the quoting worked differently on GHA hosted agent and on locally hosted agent. Sigh... Probably, my poor PS skills. In the end it made me revert to old as mammoth, but relatively reliable CMD :/

compnerd:
Pushed a bunch of style changes.  I think that we want to go further and convert the toolchain build to something that allow iteration as well, as it should be built twice (once for X64, once for ARM64).

build.ps1
```diff
@@ -76,11 +101,12 @@ function Build-CMakeProject
...
 }
 
-function Build-ToolchainX64
+function Build-Toolchain($Arch)
 {
+
   Build-CMakeProject `
     -B $BinaryCache\1 `
-    -C $SourceCache\swift\cmake\caches\Windows-x86_64.cmake `
+    -C $SourceCache\swift\cmake\caches\Windows-$Arch.LLVMName.cmake `
```
tristanlabelle, line 109: This will try to resolve `$Arch.LLVMName.cmake`, you need `$($Arch.LLVMName).cmake`. Syntax coloring helps to catch these issues. @compnerd

build.ps1
```diff
@@ -409,195 +435,234 @@ function Build-SwiftCorelibsXCTest($Arch)
...
     -D CMAKE_MT=mt `
     -G Ninja `
     -S $SourceCache\sqlite-3.36.0 `
     -Targets ($null, "install")
 }
 
-function Build-SwiftSystemX64
+function Build-SwiftSystem($Arch)
 {
+  $SwiftBuildDir = Get-ProjectBuildDir($Arch, 1)
```
tristanlabelle, line 471: Careful, function call syntax does do what you think in powershell. It will pass a tuple as the first argument. I thought `Set-StrictMode` would catch this

build.ps1
```diff
@@ -409,195 +435,234 @@ function Build-SwiftCorelibsXCTest($Arch)
...
     -G Ninja `
     -S $SourceCache\swift-system `
     -Targets ($null, "install")
 }
 
-function Build-ToolsSupportCoreX64
+function Build-ToolsSupportCore($Arch)
 {
+  $ArchName = $Arch.ShortName
+  $SwiftBuildDir = Get-ProjectBuildDir($Arch, 1)
```
tristanlabelle, line 490: Same, function call syntax is incorrect

build.ps1
```diff
@@ -673,61 +748,38 @@ function Build-SourceKitLspX64
...
-Build-LibXml2 $ArchX64
-Build-Curl $ArchX64
-Build-Icu $ArchX64
-Build-SwiftRuntime $ArchX64
-Build-SwiftCorelibsLibdispatch $ArchX64
-Build-SwiftCorelibsFoundation $ArchX64
-Build-SwiftCorelibsXCTest $ArchX64
-
-if (-not $OnlyX64)
+foreach ($Arch in $ArchX64,$ArchX86,$ArchARM64)
```
tristanlabelle, line 756: It would be nice to keep supporting `-OnlyX64` or an equivalent for local builds. In fact I'd vote for requiring `-X86 -X64 -Arm64` for building all platforms and putting the burden of passing the args on the CI automation.

tristanlabelle (commented):
@compnerd I can't request changes on my own PR but there's a few bugs in the latest revision

build.ps1
compnerd: `-OnlyX64` is too ambiguous as are the switches you suggest.  The default should be to assume that you will build all the architectures, because there are differences in between them, and that is a requirement for development.  That is just the cost of working on the toolchain.

Eventually, I would be okay with `-hosts X64 -sdks X64` as an option, with the default being `-hosts X64;ARM64 -sdks X64;X86;ARM64` and possibly restoration of the ARM SDK as well.

build.ps1
```diff
@@ -0,0 +1,923 @@
...
+{
+  $ArchName = $Arch.ShortName
+
+  if (-not(Test-Path -Path "$SourceCache\icu\icu4c\CMakeLists.txt"))
+  {
+    Copy-Item $SourceCache\swift-installer-scripts\shared\ICU\CMakeLists.txt $SourceCache\icu\icu4c\
+    Copy-Item $SourceCache\swift-installer-scripts\shared\ICU\icupkg.inc.cmake $SourceCache\icu\icu4c\
+  }
+
+  if ($Arch -eq $ArchARM64)
```
compnerd, line 284: We should compare this to `%PROCESSOR_ARCHITECTURE%`.  If we are building for a foreign architecture, we need to use the host tools.

build.ps1
```diff
@@ -0,0 +1,929 @@
...
+Build-PackageManager $HostArch
+Build-IndexStoreDB $HostArch
+Build-Syntax $HostArch
+Build-SourceKitLSP $HostArch
+
+# Switch to swift-driver
+Copy-Item -Force $BinaryCache\7\bin\swift-driver.exe $ToolchainInstallRoot\usr\bin\swift.exe
+Copy-Item -Force $BinaryCache\7\bin\swift-driver.exe $ToolchainInstallRoot\usr\bin\swiftc.exe
+
+$python = "${Env:ProgramFiles(x86)}\Microsoft Visual Studio\Shared\Python39_64\python.exe"
```
compnerd, line 915: It might make sense to prefer the `python` in `Path` and then fallback to this.  This is also now out of support ...

tristanlabelle: Agreed. Also I suspect python only get installed there by older version of VS because mine didn't even after checking the python individual component

compnerd:
Lets see what happens ...

PR #458, opened 2023-03-07
GHA,Azure: begin migration towards unadulterated Visual Studio

compnerd:
Begin the process of altering the build configuration to use the SDK generated VFS overlay.  The toolchain migration is a bit more complicated as the VFS overlay does not exist and needs to be generated.


compnerd:
@lxbndr has taken care of this graciously!

PR #459, opened 2023-03-08
build: various tweaks

compnerd:
- Centralise `-D CMAKE_M=mt` as that is a CMake workaround
- Clean up bleeding whitespace
- Avoid installing static Swift packages
- Support multiple python.exe in Path

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -123,10 +123,11 @@ function Build-CMakeProject
     Invoke-VsDevShell $Arch
     $CurrentVSDevShellTargetArch = $Arch
   }
-  
+
   # Add additional defines (unless already present)
   $Defines = $Defines.Clone()
   TryAddKeyValue $Defines CMAKE_BUILD_TYPE $BuildType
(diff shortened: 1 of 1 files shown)


build.ps1
```diff
@@ -947,7 +927,7 @@ Copy-Item -Force $BinaryCache\7\bin\swift-driver.exe $ToolchainInstallRoot\usr\b
 $python = "${Env:ProgramFiles(x86)}\Microsoft Visual Studio\Shared\Python39_64\python.exe"
 if (-not (Test-Path $python))
 {
-  $python = ((where.exe python) | Out-String).Trim()
+  $python = ((where.exe python) | Select -Frist 1 | Out-String).Trim()
```
tristanlabelle, line 930: Frist!!1!
```suggestion
  $python = (where.exe python) | Select -First 1
```

compnerd: I typo real gud!

PR #461, opened 2023-03-08
Add SDKs param to script allowing building a subset of architectures

tristanlabelle:
This allows calling the script with `build -SDKs X64,X86` to build only for a subset of architectures.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -2,7 +2,7 @@
 # Copyright 2023 Tristan Labelle <tristan@thebrowser.company>
 
 param(
-  [string[]] $Archs = @("X64","X86","Arm64")
+  [string[]] $SDKs = @("X64","X86","Arm64")
 )
 
 $ErrorActionPreference = "Stop"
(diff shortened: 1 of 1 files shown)


build.ps1
```diff
@@ -1,6 +1,10 @@
 # Copyright 2020 Saleem Abdulrasool <compnerd@compnerd.org>
 # Copyright 2023 Tristan Labelle <tristan@thebrowser.company>
 
+param(
+  [string[]] $Archs = @("X64","X86","Arm64")
```
compnerd, line 5: Please name this `SDKs`.  It is ambiguous for the architecture for the toolchain vs SDK.  Thoughts on `ARM64` rather than `Arm64`?

compnerd (changes requested):

build.ps1
tristanlabelle: Done in latest. String comparisons are case insensitive in Powershell, even the switch statement, I verified

compnerd (approved):
This looks great!

PR #462, opened 2023-03-08
build: install the SDK modulemaps

compnerd:
Add the SDK modulemaps into the SDK as this is now needed at runtime.


tristanlabelle (approved):

compnerd:
Let's not do this: https://github.com/apple/swift/pull/64218 is the better way to handle this.  The more we can avoid staging in `build.ps1` the better.

PR #464, opened 2023-03-09
GHA,Azure: alter terminology for upcoming restructuring

compnerd:
Prepare for some restructuring of the toolchain distribution to allow us to package things better for redistribution and installation purposes.


compnerd:
Most of the renaming here has transitioned

PR #468, opened 2023-03-10
Add $Arch parameter to Build-WiXProject

tristanlabelle:
Supports specifying which of the `wxs` files should be built on a given invocation.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -254,17 +254,21 @@ function Build-WiXProject()
 {
   [CmdletBinding(PositionalBinding = $false)]
   param(
-    [Parameter(Position = 0)]
+    [Parameter(Position = 0, Mandatory = $true)]
     [string]$FileName,
+    [Parameter(Mandatory = $true)]
+    [hashtable]$Arch,
     [hashtable]$Properties = @{}
   )
 
   $Name = $FileName.Split('.')[0]
+  $ArchName = $Arch.VSName
 
   $Properties = $Properties.Clone()
+  TryAdd-KeyValue $Properties ProductArchitecture $ArchName
   TryAdd-KeyValue $Properties RunWixToolsOutOfProc true
-  TryAdd-KeyValue $Properties OutputPath $BinaryCache\msi\
-  TryAdd-KeyValue $Properties IntermediateOutputPath BinaryCache\$Name\
+  TryAdd-KeyValue $Properties OutputPath $BinaryCache\msi\$ArchName\
+  TryAdd-KeyValue $Properties IntermediateOutputPath BinaryCache\$Name\$ArchName\
 
   $MSBuildArgs = @("$SourceCache\swift-installer-scripts\platforms\Windows\$FileName")
   $MSBuildArgs += "-noLogo"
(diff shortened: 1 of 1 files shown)


build.ps1
```diff
@@ -254,17 +254,21 @@ function Build-WiXProject()
...
   $Name = $FileName.Split('.')[0]
+  $ArchName = $Arch.VSName
 
   $Properties = $Properties.Clone()
+  TryAdd-KeyValue $Properties ProductArchitecture $ArchName
   TryAdd-KeyValue $Properties RunWixToolsOutOfProc true
-  TryAdd-KeyValue $Properties OutputPath $BinaryCache\msi\
-  TryAdd-KeyValue $Properties IntermediateOutputPath BinaryCache\$Name\
+  TryAdd-KeyValue $Properties OutputPath $BinaryCache\msi\$ArchName\
+  TryAdd-KeyValue $Properties IntermediateOutputPath BinaryCache\$Name\$ArchName\
```
tristanlabelle, line 271: Alternative output path suggestions are welcome

build.ps1
```diff
@@ -254,17 +254,21 @@ function Build-WiXProject()
...
+    [Parameter(Mandatory = $true)]
+    [hashtable]$Arch,
     [hashtable]$Properties = @{}
   )
 
   $Name = $FileName.Split('.')[0]
+  $ArchName = $Arch.VSName
 
   $Properties = $Properties.Clone()
+  TryAdd-KeyValue $Properties ProductArchitecture $ArchName
```
compnerd, line 268: This is worrisome.  The SDKs are architecture agnostic (they are basically headers and import libraries).

build.ps1
```diff
@@ -933,28 +937,37 @@ function Build-SourceKitLSP($Arch)
...
-  #   SDK_ROOT = "$SDKInstallRoot\";
-  #   SWIFT_SOURCE_DIR = "$SourceCache\swift\";
-  # }
+  foreach ($Arch in $SDKArchs)
+  {
+    # WiX v3 does not support ARM64
+    if ($Arch -eq $ArchARM64)
+    {
+      continue
+    }
```
compnerd, line 951: We still can package the ARM64 SDK as it is architecture agnostic.

build.ps1
tristanlabelle: But we have `sdk-x86.wxs`, `sdk-amd64.wxs` and `sdk-arm64.wxs` which are selected in `sdk.wixproj` using:

```xml
  <ItemGroup>
    <Compile Include="sdk-$(ProductArchitecture).wxs" />
  </ItemGroup>

Comment (compnerd) [reply build.ps1:268]: Ah, that key is different from the actual key itself of what the architecture is that the MSI supports!

File: build.ps1 (hunk 3 above)
Comment (compnerd) [reply build.ps1:951]: Hmm, plausibly.  I thought I had packaged up the ARM64 SDK previously for installation on X64.

File: build.ps1 [hunk 4]
```diff
@@ -933,28 +937,37 @@ function Build-SourceKitLSP($Arch)
...
 
-  Build-WiXProject runtime.wixproj -Properties @{
-    SDK_ROOT = (Get-RuntimeInstallDir $ArchX64) + "\";
+    Build-WiXProject runtime.wixproj -Arch $Arch -Properties @{
+      SDK_ROOT = (Get-RuntimeInstallDir $ArchX64) + "\";
```
compnerd, line 954: That should be `$Arch` not `$ArchX64` as the runtimes are architecture dependent.

tristanlabelle: Good call. Fixed in latest

build.ps1
tristanlabelle: I didn't understand? It seems like `ProductArchitecture` is the only knob we have to choosing which msi we're building:

```
    <ProductArchitecture Condition=" '$(ProductArchitecture)' == '' ">amd64</ProductArchitecture>
    <ProductArchitecture>$(ProductArchitecture)</ProductArchitecture>
```

```
  <PropertyGroup Condition=" '$(ProductArchitecture)' == 'amd64' ">
    <InstallerPlatform>x64</InstallerPlatform>
  </PropertyGroup>
  <PropertyGroup Condition=" '$(ProductArchitecture)' == 'arm64' ">
    <InstallerPlatform>arm64</InstallerPlatform>
    <InstallerVersion>500</InstallerVersion>
  </PropertyGroup>
  <PropertyGroup Condition=" '$(ProductArchitecture)' == 'arm' ">
    <InstallerPlatform>arm</InstallerPlatform>
    <InstallerVersion>500</InstallerVersion>
  </PropertyGroup>
  <PropertyGroup Condition=" '$(ProductArchitecture)' == 'x86' ">
    <InstallerPlatform>x86</InstallerPlatform>
  </PropertyGroup>

Comment (compnerd) [reply build.ps1:268]: The `ProductArchitecture` indicates the architecture, but we don't have a `<Condition>` that prevents the installation on another architecture.

File: build.ps1 (hunk 1 above)
Comment (compnerd) [reply build.ps1:271]: 🤷‍♂️ Seems like a reasonable thing for now.  we could also flip the two if we wanted, so `$BinaryCache\$ArchName\msi`.  That would isolate it based on the architecture.  I don't feel too strongly about this, its internals and we can restructure this as/when needed.

File: build.ps1 (hunk 3 above)
Comment (tristanlabelle) [reply build.ps1:951]:
```
candle.exe : error CNDL0264: The parameter 'arch' is missing or has an invalid value arm64.  Possible values are x86, x64, or ia64. [S:\SourceCache\swift-installer-scripts\platforms\Windows\runtime.wixproj]

compnerd: Might've used a newer snapshot of WiX.  Migrating to 4.0 is going to require some work.

compnerd (approved):

PR #469, opened 2023-03-13
build: use the staged SDK for the build

compnerd:
This avoids the use of `-vfsoverlay` with a pointer into the Swift build tree.  This should allow us to use the same technique during the build on the CI systems.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -146,6 +146,7 @@ function Build-CMakeProject
     [string] $CacheScript = "",
     [string[]] $UseMSVCCompilers = @(), # C,CXX
     [string[]] $UseBuiltCompilers = @(), # ASM,C,CXX,Swift
+    [string] $SwiftSDK = "",
     [hashtable] $Defines = @{},
     [switch] $BuildDefaultTarget = $false,
     [string[]] $BuildTargets = @()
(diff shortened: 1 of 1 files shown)


build.ps1
```diff
@@ -198,17 +199,27 @@ function Build-CMakeProject
...
-      "-Xlinker /INCREMENTAL:NO",
-      "-Xlinker /DEBUG",
-      "-Xlinker /OPT:REF",
-      "-Xlinker /OPT:ICF"
-    ) -Join " "
 
+    $SwiftArgs = [System.Collections.ArrayList]@()
+
+    if ($SwiftSDK -ne "") {
+      $SwiftArgs.Add("-sdk $SwiftSDK") | Out-Null
```
tristanlabelle, line 206: I believe you can use a semicolon instead of ` | Out-Null` to suppress automatic expression output.

build.ps1
```diff
@@ -198,17 +199,27 @@ function Build-CMakeProject
...
-      "-L $SwiftResourceDir\windows",
-      "-vfsoverlay $RuntimeBuildDir\stdlib\windows-vfs-overlay.yaml",
-      "-g -debug-info-format=codeview",
-      "-Xlinker /INCREMENTAL:NO",
-      "-Xlinker /DEBUG",
-      "-Xlinker /OPT:REF",
-      "-Xlinker /OPT:ICF"
-    ) -Join " "
 
+    $SwiftArgs = [System.Collections.ArrayList]@()
```
tristanlabelle, line 203: Why an explicit ArrayList?

build.ps1
tristanlabelle: If this were a regular array (not sure if ArrayList will behave the same), you could do:
`$SwftArgs += "-sdk $SwiftSDK"`
And it shouldn't print anything because it's more than a simple expression, there's an assignment

tristanlabelle (approved):

build.ps1
compnerd: The `@()` is fixed in size and causes an exception.  Am I doing something wrong?

build.ps1
compnerd: 🤯 Let me try that!

compnerd: No, that still prints the index :-(

build.ps1
tristanlabelle: When using `.Add`? That method shouldn't be part of the array interface but maybe it would be with Powershell. I'm using `+` or `+=` to concatenate arrays successfully elsewhere in this script (maybe it's reallocating but I don't think we care)

compnerd: Yes, the `+=` reallocates

PR #475, opened 2023-03-14
Specify CMAKE_Swift_COMPILER_TARGET when building the runtime

tristanlabelle:
This flag was missing when compared with the official build. See 
https://github.com/compnerd/swift-build/blob/67af2270d0b605f8c7691e32b2abf5b5bb298cec/.azure/vs2022.yml#L876

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -504,6 +504,7 @@ function Build-Runtime($Arch)
     -BuildDefaultTarget `
     -Defines @{
       CMAKE_INSTALL_BINDIR = Get-RuntimeInstallDir $Arch "usr\bin";
+      CMAKE_Swift_COMPILER_TARGET = $Arch.LLVMTarget;
       LLVM_DIR = "$LLVMBuildDir\lib\cmake\llvm";
       SWIFT_ENABLE_EXPERIMENTAL_CONCURRENCY = "YES";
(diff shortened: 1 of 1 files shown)


build.ps1
```diff
@@ -504,6 +504,7 @@ function Build-Runtime($Arch)
     -BuildDefaultTarget `
     -Defines @{
       CMAKE_INSTALL_BINDIR = Get-RuntimeInstallDir $Arch "usr\bin";
+      CMAKE_Swift_COMPILER_TARGET = $Arch.LLVMTarget;
```
compnerd, line 507: Hmm, I don't think that this matters as this is not currently honored by this build.  This _should_ generate a warning.

tristanlabelle: Checking locally but the CI build that does the same did not generate a warning:

https://github.com/thebrowsercompany/swift-build/actions/runs/4419433944/jobs/7751702142

tristanlabelle: It did not generate a warning locally

FreddyEngelkingMei:
the owner, reviews the suggestion but disagrees, saying that the flag is not currently honored by the build and that it should generate a warning. Tristanlabelle then checks locally and reports that the CI build did not generate a warning with the flag. It seems that they are trying to resolve an issue related to building the Swift runtime for Windows using the Microsoft Visual Studio 2022 compiler.

compnerd (approved):

PR #477, opened 2023-03-15
Leverage CMAKE_INSTALL_BINDIR for XCTest

tristanlabelle:
Avoids a manual copy step.

Depends on https://github.com/apple/swift-corelibs-xctest/pull/433


compnerd (approved):
Awesome, thanks!

PR #481, opened 2023-03-16
Fix dispatch headers for sdk.wixproj

tristanlabelle:
Copies dispatch headers to the installed layout location instead of moving them such that the installer can still find them where they are initially installed by cmake.

This helps unblock the installer but a better fix would be to check if we can change this:
https://github.com/apple/swift-corelibs-libdispatch/blob/e7c44163ce06a916f0641a8446e392ef28929309/CMakeLists.txt#L294

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -538,12 +538,13 @@ function Build-Dispatch($Arch)
       BUILD_TESTING = "NO";
     }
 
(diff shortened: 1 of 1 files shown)


tristanlabelle:
Made moot by https://github.com/compnerd/swift-build/pull/483

PR #483, opened 2023-03-16
Use arch-specific CMake platform install roots to unblock building sdk.wixproj

tristanlabelle:
Update `build.ps1` to use an architecture-specific cmake install platform directory tree when building, and add a consolidation step that builds the installed layout for local usage. This unblocks building `sdk.wixproj` and simplifies the logic for `runtime.wixproj` by preserving the original build tree that is expected by the installer.


tristanlabelle:
Pending testing by building the SDK for all three archs.

build.ps1
```diff
@@ -643,23 +611,70 @@ function Build-XCTest($Arch)
...
-    $InstallDir\lib\swift\windows\XCTest.swiftmodule\$($Arch.LLVMTarget).swiftdoc
-  Move-Item -Force `
-    $InstallDir\lib\swift\windows\$($Arch.LLVMName)\XCTest.swiftmodule `
-    $InstallDir\lib\swift\windows\XCTest.swiftmodule\$($Arch.LLVMTarget).swiftmodule
+# Copies files installed by CMake from the arch-specific platform root,
+# where they follow the layout expected by the installer,
+# to the final platform root, following the installer layout.
+function Consolidate-PlatformInstall($Arch)
+{
+  Remove-Item -Force -Recurse $PlatformInstallRoot -ErrorAction Ignore
```
tristanlabelle, line 634: Bug: This needs to happen once before building all architectures

build.ps1
```diff
@@ -291,6 +306,11 @@ function Build-WiXProject()
   Check-LastExitCode
 }
 
+function Write-PListFile($Path, $Key, $Value)
```
compnerd, line 309: This isn't generic, please undo this.  Alternatively, it should take a JSON blob and serialise the into a plist.

build.ps1
```diff
@@ -82,13 +94,16 @@ function Get-ProjectBuildDir($Arch, $ID)
   return "$BinaryCache\" + ($Arch.BuildID + $ID)
 }
 
+# We need to use different trees for the platform directory because the path to Swift
+# modules .lib files would clash across architectures.
+function Get-ArchSpecificPlatformDir($Arch, $SubDir = "")
+{
+  return [IO.Path]::Combine("$BinaryCache\Windows.platform\$($Arch.ShortName)", $SubDir)
```
compnerd, line 101: I'd rather have the architecture be hoisted above the platform if you want to do this.

build.ps1
```diff
@@ -643,23 +611,70 @@ function Build-XCTest($Arch)
...
+  # Copy SDK modules, restructuring to the installed layout
+  Get-ChildItem -Recurse -Filter *.swiftmodule $WindowsLibSrc\$($Arch.LLVMName) | ForEach-Object {
+    $DstDir = "$WindowsLibDst\$($_.Name)"
+    New-Item -ItemType Directory $DstDir | Out-Null
+
+    # Copy all module files, renaming to the arch-specific variant
+    Get-ChildItem $_.Directory | ForEach-Object {
+      Copy-Item -Force $_.FullName "$DstDir\$($Arch.LLVMTarget)$($_.Extension)"
+    }
+  }
```
compnerd, line 673: Doesn't this prevent us from actually using the SDK during the build?  I'd like to be able to use the SDK layout during the build and cross-compile.

compnerd (changes requested):

build.ps1
tristanlabelle: I iterated until it diffs to the same files as the installed sdk. I'm going to do a clean build overnight to confirm this

build.ps1
tristanlabelle: Bah, it was good enough for our needs. I'd need to serialize into python so I'll revert

tristanlabelle: In fact I just wanted to break the long line, so I did that with a backtick

build.ps1
tristanlabelle: Final tweaks this morning and I could use the SDK to build every other project up to and including sourcekit-lsp

build.ps1
tristanlabelle: Done

tristanlabelle:
I succeeded in building the whole shebang with the latest iteration!

build.ps1
```diff
@@ -1063,20 +1076,4 @@ Build-SourceKitLSP $HostArch
...
-    throw "Python.exe not found"
-  }
-}
-
-# SDKSettings.plist
-Invoke-Program $python -c "import plistlib; print(str(plistlib.dumps({ 'DefaultProperties': { 'DEFAULT_USE_RUNTIME': 'MD' } }), encoding='utf-8'))" > $SDKInstallRoot\SDKSettings.plist
-
-# Info.plist
-Invoke-Program $python -c "import plistlib; print(str(plistlib.dumps({ 'DefaultProperties': { 'XCTEST_VERSION': 'development' } }), encoding='utf-8'))" > $PlatformInstallRoot\Info.plist
+Copy-Item -Force $BinaryCache\7\bin\swift-driver.exe $ToolchainInstallRoot\usr\bin\swiftc.exe
```
compnerd, line 1079:
```suggestion
Copy-Item -Force $BinaryCache\7\bin\swift-driver.exe $ToolchainInstallRoot\usr\bin\swiftc.exe
```

compnerd (approved):

build.ps1
tristanlabelle: What's the change? 🤔

PR #486, opened 2023-03-17
Prevent filesystem side effects from build.ps1 -ToBatch

tristanlabelle:
It now instantly produces a cmd version of the cmake/msbuild invocations. Also adds support for redirection to nul and files.


build.ps1
```diff
@@ -105,8 +105,18 @@ function Get-ProjectBuildDir($Arch, $ID)
...
 {
+  [CmdletBinding(PositionalBinding = $false)]
+  param(
+    [Parameter(Position = 0, Mandatory = $true)]
+    [string] $Executable,
+    [switch] $OutNull = $false,
+    [string] $OutFile = "",
+    [Parameter(Position = 1, ValueFromRemainingArguments)]
+    [string[]] $Args
+  )
```
compnerd, line 118: Would you mind enlightening me as to the difference here?  Why the `CmdletBinding` rather than the parameters?

compnerd (approved):

build.ps1
tristanlabelle: By default, arguments will bind positionally and in order even without `-Name`. In the case here, I want to enable call like `Invoke-Program foo.exe myarg1 myarg2` where `foo.exe` binds to `$Executable`, `myarg myarg2` binds to `$Args`, and nothing binds to `$OutNull or $OutFile` without an explicit `-OutNull` or `-OutFile path`.

PR #487, opened 2023-03-23
Add -restore to msbuild invocations for WiX v4 compatibility

tristanlabelle:
Prepares the CI definitions for the migration of the Windows installer to WiX v4, which relies on Nuget and so require a package restore step: https://github.com/apple/swift-installer-scripts/pull/184

### DIFF
--- a/.azure/vs2022.yml
+++ b/.azure/vs2022.yml
@@ -1555,10 +1555,9 @@ stages:
               platform: $(platform)
               configuration: Release
               maximumCpuCount: true
-              restoreNugetPackages: false
+              restoreNugetPackages: true
               createLogFile: true
               msbuildArguments:
-                -restore
                 -p:RunWixToolsOutOfProc=true
                 -p:CERTIFICATE=$(certificate.secureFilePath)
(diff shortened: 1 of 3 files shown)


.azure/vs2022.yml
```diff
@@ -1558,6 +1558,7 @@ stages:
               restoreNugetPackages: false
               createLogFile: true
               msbuildArguments:
+                -restore
```
compnerd, line 1561: Can you not use `restoreNugetPackages` and set it to `true` instead?

https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/msbuild-v1?view=azure-pipelines

tristanlabelle: Oh probably. The doc says that it's deprecated but I think it'll still be good enough for us because their workaround is more complicated:

> This option is deprecated. To restore NuGet packages, add a [NuGet](https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/package/nuget) task before the build.

compnerd (approved):
Do we need to be worried about compatibility with this change?  That is, this should be a no-op before the changes for WiX 4 get merged right?

PR #489, opened 2023-03-23
build.ps1 -ProductVersion

tristanlabelle:
Support a `-ProductVersion` parameter which gets passed down when building the WiX projects.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -1,10 +1,12 @@
 # Copyright 2020 Saleem Abdulrasool <compnerd@compnerd.org>
 # Copyright 2023 Tristan Labelle <tristan@thebrowser.company>
 
+[CmdletBinding(PositionalBinding = $false)]
 param(
-  [string[]] $SDKs = @("X64","X86","Arm64"),
   [string] $SourceCache = "S:\SourceCache",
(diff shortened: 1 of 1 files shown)


build.ps1
```diff
@@ -1,10 +1,12 @@
 # Copyright 2020 Saleem Abdulrasool <compnerd@compnerd.org>
 # Copyright 2023 Tristan Labelle <tristan@thebrowser.company>
 
+[CmdletBinding(PositionalBinding = $false)]
 param(
-  [string[]] $SDKs = @("X64","X86","Arm64"),
   [string] $SourceCache = "S:\SourceCache",
   [string] $BinaryCache = "S:\b",
+  [string[]] $SDKs = @("X64","X86","Arm64"),
+  [string] $ProductVersion = "0.0.0",
```
compnerd, line 9:
```suggestion
  [string] $ProductVersion = "0.0.0.0",
```

tristanlabelle: I agree but I was staying consistent with:

```xml
<ProductVersion Condition=" '$(ProductVersion)' == '' ">0.0.0</ProductVersion>

PR #492, opened 2023-03-24
build.ps1: Consolidate the redist installation mimicking Program Files

tristanlabelle:
Updates the consolidation step to write to folders mimicking Program Files native and WoW directories.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -721,7 +721,7 @@ function Consolidate-RedistInstall($Arch)
   {
     $ProgramFilesName = "Program Files (x86)"
   }
-  elseif (($HostArch -eq $HostArm64) -and ($Arch -eq $ArchX64))
+  elseif (($HostArch -eq $ArchArm64) -and ($Arch -eq $ArchX64))
   {
     # x64 programs actually install under "Program Files" on arm64,
     # but this would conflict with the native installation.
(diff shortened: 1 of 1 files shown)


build.ps1
```diff
@@ -695,8 +692,30 @@ function Consolidate-RedistInstall($Arch)
...
+  }
+  elseif ($Arch -eq $ArchX86)
+  {
+    $ProgramFilesName = "Program Files (x86)"
+  }
+  elseif (($HostArch -eq $HostArm64) -and ($Arch -eq $ArchX64))
+  {
+    # x64 programs actually install under "Program Files" on arm64,
+    # but this would conflict with the native installation.
+    $ProgramFilesName = "Program Files (x64)"
```
compnerd, line 707: Interesting, how does Microsoft handle the conflict?  What if you need/want both versions?

compnerd, line 707:
```suggestion
    $ProgramFilesName = "Program Files (Amd64)"
```

This would match `Program Files (Arm)` better ...

build.ps1
```diff
@@ -695,8 +692,30 @@ function Consolidate-RedistInstall($Arch)
...
+  }
+  else
+  {
+    # arm64 cannot be installed on x64
```
compnerd, line 711: 😢

build.ps1
```diff
@@ -695,8 +692,30 @@ function Consolidate-RedistInstall($Arch)
...
+    return
+  }
+
+  $RedistInstallRoot = "S:\$ProgramFilesName\swift\runtime-development"
```
compnerd, line 715: I wonder if we want to use a parameter for the stage location, but that is outside the scope of this change I think.

build.ps1
```diff
@@ -695,8 +692,30 @@ function Consolidate-RedistInstall($Arch)
...
+
+  Remove-Item -Force -Recurse $RedistInstallRoot -ErrorAction Ignore
+  Copy-Dir "$($HostArch.SDKInstallRoot)\usr\bin" "$RedistInstallRoot\usr"
```
compnerd, line 718: This should be `$($Arch.SDKInstallRoot)` no?

compnerd (approved):

build.ps1
tristanlabelle: I haven't found anything official but it seems like this is not a scenario that they went through extra lengths to support.

https://stackoverflow.com/questions/74402454/what-is-each-program-files-folder-purpose-in-arm64-architecture

PR #493, opened 2023-03-24
Add consolidate toolchain step

tristanlabelle:
Build the toolchain to an arch-specific directory and add a consolidate step (with the special logic for `_InternalSwiftScan`) like the rest of the build. Also adds in `LLDB_PYTHON_EXT_SUFFIX = ".pyd";` to unblock building `toolchain.wixproj`.

See related:
- https://github.com/apple/swift-installer-scripts/pull/179
- https://github.com/apple/swift/pull/64364


build.ps1
```diff
@@ -57,6 +58,7 @@ $ArchX86 = @{
   PlatformInstallRoot = "$BinaryCache\x86\Windows.platform\";
   SDKInstallRoot = "$BinaryCache\x86\Windows.platform\Developer\SDKs\Windows.sdk";
   XCTestInstallRoot = "$BinaryCache\x86\Windows.platform\Developer\Library\XCTest-development";
+  ToolchainInstallRoot = "$BinaryCache\x86\unknown-Asserts-development.xctoolchain";
```
compnerd, line 61: This should be removed.  I do not expect to ever support a 32-bit toolchain, that will always require cross-compilation from a 64-bit toolchain.  The 32-bit targets are where you can run code only.

compnerd (approved):

build.ps1
tristanlabelle: Makes sense! I removed it in latest.

PR #497, opened 2023-03-27
build.ps1: Enable semantic versioning for the installer

tristanlabelle:
WiX v4 will enforce `Packages`' `Version` to be numerical, but will allow `Bundle`s' containing semantic versioning suffixes. It's unclear if this is by design, but it results in a way to get detailed version information in the add/remove programs control panel:

image

Depends on https://github.com/apple/swift-installer-scripts/pull/184 because this does not work with WiX v3


compnerd (approved):
Awesome!  Thanks

PR #501, opened 2023-03-29
Update REFLECTION/OBSERVATION experimental flags in Windows build

tristanlabelle:
`SWIFT_ENABLE_EXPERIMENTAL_REFLECTION` does not exist anymore, and `SWIFT_ENABLE_EXPERIMENTAL_OBSERVATION` now does. This aligns with the Apple CI build.


compnerd:
This is technically incorrect - the flag did exist but was recently removed.  The observation module is new and should be enabled.  The end result is correct, the explanation is not.

PR #502, opened 2023-03-31
Enable the arm64 installer build

tristanlabelle:
Now that install scripts use WiX v4, we can build WiX projects for arm64.


compnerd (approved):
It's so beautiful!

PR #503, opened 2023-03-31
build.ps1: Uniformize binary roots for non-numeric projects

tristanlabelle:
Introduce `$Arch.BinaryRoot` and use it as the uniform binary root for all non-numeric projects. This doesn't make paths longer (`libxml2-2.9.12.x64` vs `x64\libxml2-2.9.12`).


compnerd (approved):
TIHI.  Goodbye muscle memory.  This is much nicer.

PR #504, opened 2023-04-03
build: ensure that we write out the content in UTF-8

compnerd:
We expect the content to be UTF-8, explicitly specify the encoding for the files that are being created.  This is important to enable the use of VSCode.


tristanlabelle (approved):
`Write-Output -Encoding UTF8 $OutputLine` won't always do the right thing on Windows but this is a gigantic can of worms that consumed many months of my life so I'll look the other way 😅. The `Out-File` one is the important one and that's okay

tristanlabelle (approved):

PR #505, opened 2023-04-04
build.ps1: Support staging artifacts as expected by the Apple CI

tristanlabelle:
Adds an optional step to copy the artifacts (msi's and installer) to where the Apple CI will expect them.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -8,6 +8,7 @@ param(
   [string[]] $SDKs = @("X64","X86","Arm64"),
   [string] $ProductVersion = "0.0.0",
   [string[]] $Test = @(),
+  [string] $StageArtifactsTo = "",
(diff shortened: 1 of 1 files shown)


build.ps1
```diff
@@ -8,6 +8,7 @@ param(
   [string[]] $SDKs = @("X64","X86","Arm64"),
   [string] $ProductVersion = "0.0.0",
   [string[]] $Test = @(),
+  [string] $StageArtifactsTo = "",
```
compnerd, line 11: How about `Stage` instead of `StageArtifactsTo`?  That makes the usage much simpler.

tristanlabelle: Good idea, done!

compnerd (approved):

PR #512, opened 2023-04-06
build.ps1: Work around limitations in passing arrays from cmd

tristanlabelle:
From PowerShell, this will work: `& build.ps1 -Test foo,bar` and be interpreted as `$Test = @("foo", "bar")`. However, from cmd, this: `powershell.exe -File build.ps1 -Test foo,bar` will be interpreted as `$Test = @("foo,bar")` and there is no way to disambiguate and pass in a two-entry array. Since we don't expect commas in our array element names, we can work around this in PowerShell by splitting on the comma ourselves.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -86,6 +86,10 @@ if (-not (Test-Path $python))
   }
 }
 
+# Work around limitations of cmd passing in arrays to PowerShell
+if ($SDKs.Length -eq 1) { $SDKs = $SDKs[0].Split(",") }
+if ($Test.Length -eq 1) { $Test = $Test[0].Split(",") }
+
 if ($Test -contains "*")
 {
   $Test = @("swift", "dispatch", "foundation", "xctest")
--- a/build.ps1
+++ b/build.ps1
@@ -86,7 +86,7 @@ if (-not (Test-Path $python))
   }
 }
 
-# Work around limitations of cmd passing in arrays to PowerShell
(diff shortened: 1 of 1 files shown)


build.ps1
```diff
@@ -86,6 +86,10 @@ if (-not (Test-Path $python))
   }
 }
 
+# Work around limitations of cmd passing in array arguments via powershell.exe -File
+if ($SDKs.Length -eq 1) { $SDKs = $SDKs[0].Split(",") }
+if ($Test.Length -eq 1) { $Test = $Test[0].Split(",") }
```
compnerd, line 91: Hahaha, I should introduce a project with comma in the name 😈

compnerd (approved):

tristanlabelle:
@compnerd The reason for the comma is to be consistent with what powershell recognizes as syntax when calling the script from within powershell. Semicolon would end the powershell statement:

```
Write-Output "-Test $Test ($($Test.Length))"
exit

PS S:\SourceCache\swift-build> & .\build.ps1 -Test foo,bar
-Test foo bar (2)
PS S:\SourceCache\swift-build> & .\build.ps1 -Test foo;bar
-Test foo (1)
bar : The term 'bar' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the
spelling of the name, or if a path was included, verify that the path is correct and try again.

PR #513, opened 2023-04-06
build.ps1: Support invocation in a VS developer command prompt

tristanlabelle:
The current Apple CI won't give us a choice, so we need to support running when already in a VS developer command prompt.


tristanlabelle:
Abandon in favor of patching the Apple CI to undo the vsdevcmd invocation.

PR #517, opened 2023-04-07
build.ps1: Install libs under $BinaryCache

tristanlabelle:
The Apple CI won't allow writing to `S:\Library`, so we should not use this path at all when invoked with `-SkipInstall`. The only directory we can reliably write to is `$BinaryCache`.

- Change `zlib`, `libxml2`, `sqllite`, `curl` and friends to install to `$Arch.BinaryRoot\<name>\usr` instead of under `S:\Library`. 
- Moved `$ToolchainInstallRoot`, `$PlatformInstallRoot`, `$SDKInstallRoot` to their corresponding `Install-***` functions since they are not and should not be used elsewhere.

We could now easily lift `$InstallRoot` to a parameter if needed.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -74,9 +74,6 @@ $ErrorActionPreference = "Stop"
 Set-StrictMode -Version 3.0
 
 $InstallRoot = "S:\Library"
-$ToolchainInstallRoot = "$InstallRoot\Developer\Toolchains\unknown-Asserts-development.xctoolchain"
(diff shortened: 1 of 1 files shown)


tristanlabelle:
Saleem clarified that docs for several projects are written assuming that libraries are going to be under `S:\Library`, so I will rather support redirecting the `S:\Library` folder with `-InstallRoot`

PR #527, opened 2023-04-21
Fix libxml2-utils dependency fetch

mangini:
apt-get cache is outdated, pointing to a version of
libxml2-utils that is not available anymore on ubuntu servers.

This PR runs `apt-get update` before `apt-get install`, which
ensures it is always up to date.

### DIFF
--- a/.github/workflows/swift-toolchain.yml
+++ b/.github/workflows/swift-toolchain.yml
@@ -62,6 +62,7 @@ jobs:
         name: Generate Build Context
         run: |
           # TODO(compnerd) can we make this more silent?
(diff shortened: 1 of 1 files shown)


.github/workflows/swift-toolchain.yml
```diff
@@ -62,6 +62,7 @@ jobs:
         name: Generate Build Context
         run: |
           # TODO(compnerd) can we make this more silent?
+          sudo DEBIAN_FRONTEND=noninteractive apt-get -qq update -yq
```
compnerd, line 65: This makes me sad.  Thank you for fixing this!

compnerd (approved):

PR #531, opened 2023-04-27
Uniformize CI Python versions to 3.9

tristanlabelle:


compnerd (approved):
This is the right thing to do - we just need to make sure that the next bump to `winget` does the right thing.  🤞

PR #544, opened 2023-05-24
Add workflow to build SwiftFormat.

hjyamauchi:


.github/workflows/SwiftFormat.yml
```diff
@@ -0,0 +1,76 @@
...
+          repository: nicklockwood/SwiftFormat
+
+      - uses: compnerd/gha-setup-swift@main
+        with:
+          branch: swift-5.8-release
+          tag: 5.8-RELEASE
+
+      - name: test
+        run: |
+          swift test -Xcc -IC:\Library\Developer\Toolchains\unknown-Asserts-development.xctoolchain\usr\include -Xlinker -LC:\Library\Developer\Toolchains\unknown-Asserts-development.xctoolchain\usr\lib
```
compnerd, line 31:
```suggestion
          swift test
```

Do we really need any of the additional include and library search paths?

.github/workflows/SwiftFormat.yml
```diff
@@ -0,0 +1,76 @@
...
+
+      - name: build
+        run: |
+          swift build -c release -Xcc -IC:\Library\Developer\Toolchains\unknown-Asserts-development.xctoolchain\usr\include -Xlinker -LC:\Library\Developer\Toolchains\unknown-Asserts-development.xctoolchain\usr\lib
```
compnerd, line 35:
```suggestion
          swift build -c release
```

installer-scripts/SwiftFormat-amd64.wxs
```diff
@@ -0,0 +1,54 @@
+﻿<Wix xmlns="http://wixtoolset.org/schemas/v4/wxs" xmlns:ui="http://wixtoolset.org/schemas/v4/wxs/ui">
+  <Package
+      Language="1033"
+      Manufacturer="https://github.com/nicklockwood/SwiftFormat"
```
compnerd, line 4: I think that this should be `nicklockwood` rather than the URL.

installer-scripts/SwiftFormat-amd64.wxs
```diff
@@ -0,0 +1,54 @@
...
+    </Feature>
+
+    <UI>
+      <ui:WixUI Id="WixUI_InstallDir" />
+      <Publish Dialog="WelcomeDlg" Control="Next" Event="NewDialog" Value="InstallDirDlg" Order="2" />
+      <Publish Dialog="InstallDirDlg" Control="Back" Event="NewDialog" Value="WelcomeDlg" Order="2" />
+    </UI>
+    <Property Id="WIXUI_INSTALLDIR" Value="INSTALLDIR"></Property>
+    <!-- <WixVariable Id="WixUIDialogBmp" Value="Resources\swift_dialog.png" /> -->
+    <!-- <WixVariable Id="WixUIBannerBmp" Value="Resources\swift_banner.png" /> -->
```
compnerd, line 51:
```suggestion
```

installer-scripts/SwiftFormat-arm64.wxs
```diff
@@ -0,0 +1,54 @@
...
+    </Feature>
+
+    <UI>
+      <ui:WixUI Id="WixUI_InstallDir" />
+      <Publish Dialog="WelcomeDlg" Control="Next" Event="NewDialog" Value="InstallDirDlg" Order="2" />
+      <Publish Dialog="InstallDirDlg" Control="Back" Event="NewDialog" Value="WelcomeDlg" Order="2" />
+    </UI>
+    <Property Id="WIXUI_INSTALLDIR" Value="INSTALLDIR"></Property>
+    <!-- <WixVariable Id="WixUIDialogBmp" Value="Resources\swift_dialog.png" /> -->
+    <!-- <WixVariable Id="WixUIBannerBmp" Value="Resources\swift_banner.png" /> -->
```
compnerd, line 51:
```suggestion
```

compnerd (commented):
@tristanlabelle I think that this would a good time to investigate the GUID handling.

installer-scripts/SwiftFormat-arm64.wxs
hjyamauchi: Done

installer-scripts/SwiftFormat-amd64.wxs
hjyamauchi: Done

.github/workflows/SwiftFormat.yml
hjyamauchi: Done

.github/workflows/SwiftFormat.yml
hjyamauchi: Done

installer-scripts/SwiftFormat-amd64.wxs
hjyamauchi: Done

installer-scripts/SwiftFormat-amd64.wxs
```diff
@@ -0,0 +1,52 @@
...
+            <Directory Id="Tools" Name="Tools">
+            </Directory>
+          </Directory>
+        </Directory>
+      </Directory>
+    </Directory>
+
+    <!-- Components -->
+    <ComponentGroup Id="SwiftFormat">
+      <Component Id="swiftformat.exe" Directory="Tools" Guid="77126634-5f91-40a7-b344-035ce99ef46f">
```
tristanlabelle, line 30: This would be a good opportunity to try removing Component-specific guids. I believe `ComponentGuidGenerationSeed` enables all of these guids to be generated deterministically. Full discussion here:

https://github.com/orgs/wixtoolset/discussions/7337

```suggestion
      <Component Id="swiftformat.exe" Directory="Tools">
```

.github/workflows/SwiftFormat.yml
```diff
@@ -0,0 +1,76 @@
...
+
+on:
+  workflow_dispatch:
+
+jobs:
+  windows:
+    runs-on: windows-latest
+
+    strategy:
+      matrix:
```
tristanlabelle, line 11: I'm not super familiar with github actions but do we need the matrix here if we'll be building the same tag/branch on all architectures of the build?

mangini: It is originally in @compnerd action because the plan is that this is built for an arbitrary number of toolchain releases when (a future version of) this is upstreamed to nick's SwiftFormat repo. Right now, the matrix is no-op.

tristanlabelle: Makes sense!

installer-scripts/SwiftFormat-amd64.wxs
hjyamauchi: @tristanlabelle I'm getting this error after dropping `Guid=...`

`  D:\a\swift-build\swift-build\SourceCache\installer-scripts\SwiftFormat.wxs(35): error WIX0230: The Component/@Guid attribute's value '*' is not valid for this component because it does not meet the criteria for having an automatically generated guid. Components using a Directory as a KeyPath or containing ODBCDataSource child elements cannot use an automatically generated guid. Make sure your component doesn't have a Directory as the KeyPath and move any ODBCDataSource child elements to components with explicit component guids. [D:\a\swift-build\swift-build\SourceCache\installer-scripts\SwiftFormat.wixproj]`

`Component` indeed has an a `Directory` attribute:

`      <Component Id="swiftformat.exe" Directory="Tools">`

.github/workflows/SwiftFormat.yml
compnerd: The matrix I was using was for the Swift compiler version.  Since we are not ABI stable on Windows, we need the release version matrix for the compile still as the standard library will change.

compnerd: @mangini did we test building with the 5.7 toolchain and running against a 5.9 toolchain?  I'm worried about the ABI instability in the runtime meaning that we won't be able to use the releases with newer toolchains.

installer-scripts/SwiftFormat-amd64.wxs
compnerd: I wonder if we should just merge the directory layout and component bits.  That would allow us to remove the `Directory` attribute right?

.github/workflows/SwiftFormat.yml
hjyamauchi: @compnerd Do you mean that this SwiftFormat binary is dynamically-linked and that its version needs to include the version of the swift compiler that compiled it? If so, I interpret it as that the matrix needs to have both the compiler version and the SwiftFormat version.

compnerd: @hjyamauchi correct.  `swiftCore`, `Foundation`, and `dispatch` are always dynamically linked, and the version needs to be the version from the runtime distributed with the toolchain that it was built with.  Extending the matrix to both sounds right.

installer-scripts/SwiftFormat-amd64.wxs
tristanlabelle: Ah if that syntax is valid then that would make sense! I was wondering what was the alternative to what we are doing.

hjyamauchi:
Will handle the wix improvement in a follow up PR and go ahead with this PR in the interest of making SwiftFormat available sooner.

installer-scripts/SwiftFormat.wixproj
```diff
@@ -0,0 +1,32 @@
...
+  <PropertyGroup>
+    <DefineConstants>ProductVersion=$(ProductVersion);SWIFTFORMAT_BUILD=$(SWIFTFORMAT_BUILD)</DefineConstants>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="WixToolset.UI.wixext" Version="4.0.0" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Compile Include="SwiftFormat-$(ProductArchitecture).wxs" />
```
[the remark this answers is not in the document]
tristanlabelle: I think the wixtoolset maintainer also said that component GUIDs were not system-global so we could consolidate architectures even if we can't figure out how to remove the GUIDs.

compnerd:
Merging to test and get a build

installer-scripts/SwiftFormat-amd64.wxs
hjyamauchi: @compnerd Any examples or reference on how to merge directory/component? Apple's wsx files follow this unmerged pattern.

compnerd: @hjyamauchi the wxs files there were authored by me and originally in this repository :)

I think that the `ComponentGroup` will need to be replaced by `Component`s as per the WiX Schema: https://wixtoolset.org/docs/schema/wxs/directory/

hjyamauchi: Ok I think I found a way to avoid guid in the .exe component by removing the component group and moving the component into a directory. But it still complains about the EnvironmentVariables component's missing guid when it's moved into a directory.

hjyamauchi: According to this: "Generatable guids are supported only for components with a single file as the component's keypath or no files and a registry value as the keypath." It sound slike the .exe component can have generated guid but not the environment variable one because the latter doesn't have a file or registery value as a child in this case.

hjyamauchi: https://github.com/compnerd/swift-build/pull/549 is what I got far. @compnerd @tristanlabelle Do you see a way to use autogenerated GUID for the environment variable component?

tristanlabelle: I don't but it's probably no big deal? I think the guids were local to the installer and even if they were not, it might make sense for the guids to clash if you tried to run multiple installers setting the same environment variable. (not sure)

PR #545, opened 2023-05-25
Add the swift compiler version to the matrix in the SwiftFormat yaml.

hjyamauchi:


compnerd (commented):
Thanks!  This looks like a step in the right direction.  I wonder if we should make the matrix the compiler versions and make the version of SwiftFormat an input.

.github/workflows/SwiftFormat.yml
```diff
@@ -56,8 +58,8 @@ jobs:
         with:
           draft: true
           prerelease: true
-          release_name: SwiftFormat-${{ matrix.tag }}
-          tag_name: SwiftFormat-${{ matrix.tag }}
+          release_name: SwiftFormat-${{ matrix.tag }}-${{ matrix.swift_tag }}
```
mangini, line 61: The release name will be:
`SwiftFormat-0.51.10-DEVELOPMENT-SNAPSHOT-2023-05-22-a`

Which can be a bit confusing to parse. What about:
`SwiftFormat-${{ matrix.tag }}_for_swift_toolchain_${{ matrix.swift_tag }}`

Which would become, for example:
`SwiftFormat-0.51.10_for_swift_toolchain_DEVELOPMENT-SNAPSHOT-2023-05-22-a`

compnerd: Can we do `SwiftFormat-0.51.10 (Swift 5.8)` or `SwiftFormat-0.51.10 (Swift development)`?

mangini:
> Thanks! This looks like a step in the right direction. I wonder if we should make the matrix the compiler versions and make the version of SwiftFormat an input.

This is actually what we have discussed initially. I like it.

.github/workflows/SwiftFormat.yml
compnerd, line 61:
```suggestion
          release_name: SwiftFormat-${{ matrix.tag }} (${{ matrix.swift_branch }})
```

mangini: > Can we do `SwiftFormat-0.51.10 (Swift 5.8)` or `SwiftFormat-0.51.10 (Swift development)`?

Yes! I think the release name can have parenthesis and spaces, if so, it is much better :-)

compnerd (approved):

PR #546, opened 2023-05-25
GHA: update some action versions

compnerd:
Update the version of the actions for SwiftFormat workflow to reduce the amount of warnings due to the use of deprecated node.js versions.


compnerd:
CC: @hjyamauchi @mangini

mangini (approved):

PR #547, opened 2023-05-25
GHA: update some action version

compnerd:
Update the actions versions to reduce some of the deprecation messages.


compnerd:
CC: @hjyamauchi @mangini

PR #553, opened 2023-05-31
GHA: correct a typo in the tag

compnerd:
Correct a typo that is responsible for applying the tag to the build version information.


compnerd:
CC: @tristanlabelle @mangini

PR #555, opened 2023-06-01
Add workaround for MAX_PATH with object output directories.

tristanlabelle:
`40` chars should give us plenty of space for output filenames! (famous last words)


compnerd (approved):
🤮

PR #557, opened 2023-06-01
Build swift-inspect in the toolchain workflow.

hjyamauchi:
Build swift-inspect under the installer setup at the end like swift-format.

### DIFF
--- a/.github/workflows/swift-toolchain.yml
+++ b/.github/workflows/swift-toolchain.yml
@@ -2019,9 +2019,93 @@ jobs:
           name: swift-format.exe
           path: ${{ github.workspace }}/SourceCache/swift-format/.build/x86_64-unknown-windows-msvc/release/swift-format.exe
 
+  swift_inspect:
+    runs-on: windows-latest
+    needs: [context, installer]
+
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - arch: 'amd64'
+            cpu: 'x86_64'
+            triple: 'x86_64-unknown-windows-msvc'
+
+    steps:
+      - uses: actions/download-artifact@v3
+        with:
+          name: installer-amd64
+          path: ${{ github.workspace }}/tmp
+
+      # TODO(compnerd) can this be done via a re-usage workflow for swift-format?
+
+      # TODO(compnerd): migrate this to compnerd/gha-setup-swift after the work that @mangini is doing is completed
+      - run: |
+          function Update-EnvironmentVariables {
+            foreach ($level in "Machine", "User") {
+              [Environment]::GetEnvironmentVariables($level).GetEnumerator() | % {
+                # For Path variables, append the new values, if they're not already in there
+                if ($_.Name -Match 'Path$') {
+                  $_.Value = ($((Get-Content "Env:$($_.Name)") + ";$($_.Value)") -Split ';' | Select -Unique) -Join ';'
+                }
+                $_
+              } | Set-Content -Path { "Env:$($_.Name)" }
+            }
+          }
+          try {
+            Write-Host "Starting Install installer.exe..."
+            $Process = Start-Process -FilePath ${{ github.workspace }}/tmp/installer.exe -ArgumentList ("-q") -Wait -PassThru
+            $ExitCode = $Process.ExitCode
+            if ($ExitCode -eq 0 -or $ExitCode -eq 3010) {
+              Write-Host "Installation successful"
+            } else {
+              Write-Host "non-zero exit code returned by the installation process: $ExitCode"
+              exit $ExitCode
+            }
+          } catch {
+            Write-Host "Failed to install: $($_.Exception.Message)"
+            exit 1
+          }
+          Update-EnvironmentVariables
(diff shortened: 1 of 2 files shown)


hjyamauchi:
The standalone workflow has been tested https://github.com/hjyamauchi/swift-build/actions/runs/5149995509
The toolchain workflow is being tested https://github.com/thebrowsercompany/swift-build/actions/runs/5150014765 but it may time out.

.github/workflows/swift-toolchain.yml
```diff
@@ -594,7 +594,7 @@ jobs:
             ${{ github.workspace }}/BinaryCache/1/**/*.pdb
 
       - uses: microsoft/action-publish-symbols@v2.1.6
-        if: ${{ needs.context.outputs.debug_info }}
+        if: false # ${{ needs.context.outputs.debug_info }}
```
compnerd, line 597: Please restore this.

installer-scripts/swift-inspect.wixproj
```diff
@@ -0,0 +1,32 @@
+<Project Sdk="WixToolset.Sdk/4.0.0">
```
compnerd, line 1: I think that we should consider upstreaming this into apple/swift-installer-scripts as this is something that will be in the official toolchain releases.  If this is not in the official repos we cannot have this in the toolchain releases.

compnerd (changes requested):

installer-scripts/swift-inspect.wixproj
hjyamauchi: Yep. Is it okay to merge this here first and then move it to apple/swift-installer-scripts? We cannot atomically commit across two repos, right? Or do we need to merge it into apple/swift-installer-scripts first and then merge this?

.github/workflows/swift-toolchain.yml

installer-scripts/swift-inspect.wixproj
compnerd: I think that since you have tested the build, lets get the changes merged into apple/swift-installer-scripts first and then use them here.

hjyamauchi: Sent https://github.com/apple/swift-installer-scripts/pull/204 for the wix files.

hjyamauchi:
The workflow worked (except that the release step failed due to the release tag conflict with the previous run, which I think is expected) at https://github.com/thebrowsercompany/swift-build/actions/runs/5159007140

.github/workflows/swift-toolchain.yml
```diff
@@ -2019,9 +2019,84 @@ jobs:
...
+
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ needs.context.outputs.swift_installer_scripts_revision }}
+          repository: apple/swift-installer-scripts
+          path: ${{ github.workspace }}/SourceCache/swift-installer-scripts
+
+      - name: Package
+        run: |
+          msbuild -nologo -restore -p:Configuration=Release -p:RunWixToolsOutOfProc=true -p:OutputPath=${{ github.workspace }}\artifacts -p:ProductVersion=${{ needs.context.outputs.swift_version }} -p:SWIFT_INSPECT_BUILD=${{ github.workspace }}\build ${{ github.workspace }}\SourceCache\swift-installer-scripts\platforms\Windows\swift-inspect.wixproj
```
compnerd, line 2088: We can probably avoid the copying into build and just grab the binary right from the SPM output directory.

.github/workflows/swift-toolchain.yml
```diff
@@ -2019,9 +2019,84 @@ jobs:
...
+
+      - uses: actions/upload-artifact@v2
```
compnerd, line 2090: Lets split this up into two uploads, that way we can have a separate artifact for each bit.

compnerd (approved):
LGTM

hjyamauchi:
OK, done. 
The standalone version was tested: https://github.com/hjyamauchi/swift-build/actions/runs/5161327228
The toolchain version is being tested: https://github.com/thebrowsercompany/swift-build/actions/runs/5161378872

compnerd:
Why does `swift_inspect` and `swift_format` show up as a single job?  I think that we should have them be separate jobs to allow them to execute in parallel.

hjyamauchi:
Not obvious to me why. They don't look depending on each other. Also it's odd that they are failing because of the .exe file not being found. I think the upload-artfact doesn't work with symlinks but it worked for swift-format in the previous runs?

hjyamauchi:
Updated to fix the .exe path.
Kicked off another run: https://github.com/thebrowsercompany/swift-build/actions/runs/5165466307

compnerd:
@hjyamauchi - mind rebasing the change?

hjyamauchi:
Rebased. Testing at https://github.com/thebrowsercompany/swift-build/actions/runs/5173064838
It looks like using strategy/matrix makes the swift-format/swift-inspect jobs separate.

compnerd (approved):

PR #561, opened 2023-06-05
Add a strategy matrix to the swift_format job.

hjyamauchi:
Also replace the hardcoded cpu/architecture names with references the matrix.


hjyamauchi:
Testing at https://github.com/thebrowsercompany/swift-build/actions/runs/5180218765

compnerd (approved):
Thanks!  This is a good ideas as we do want a ARM64 toolchain as well.

PR #562, opened 2023-06-06
build: use DWARF for the SDK builds

compnerd:
This adjusts the builds for the SDK to switch to DWARF.  This is problematic in many ways:
  - it adds code bloat
  - it slows down linking
  - it slightly reduces incremental builds
  - the debug information is not possible to be saved
  - the resultant binaries are in violation of the COFF specification

Given all the issues, this is still valuable as a change to enable easier debugging.  The official builds remain building with CodeView which means that the PDBs can be linked in parallel, we do not bloat the released binaries, while we get a better debugging experience locally.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -389,11 +389,13 @@ function Build-CMakeProject
   if ($UseBuiltCompilers.Contains("C")) {
     TryAdd-KeyValue $Defines CMAKE_C_COMPILER "$BinaryCache\1\bin\clang-cl.exe"
     TryAdd-KeyValue $Defines CMAKE_C_COMPILER_TARGET $Arch.LLVMTarget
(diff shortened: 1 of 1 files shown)


compnerd:
CC: @tristanlabelle @hjyamauchi

tristanlabelle (approved):

hjyamauchi:
How does this change debugging easier? Still windbg works after this?

compnerd:
> How does this change debugging easier? Still windbg works after this?

No, in fact, it actively regresses that intentionally.  It makes it easier to debug Foundation, XCTest with lldb.  WinDBG doesn't handle the DWARF, but LLDB does, and it also does make it possible to view the local variables.  The next thing to consider is trying to enable debug builds.

PR #563, opened 2023-06-06
swift-format packaging

compnerd:


compnerd:
This is taken care of already thanks to @lxbndr!

PR #565, opened 2023-06-07
Update the path string in the using the toolchain section.

hjyamauchi:
S:\Library\Developer\Platforms\Windows.platform\Developer\SDKs\Windows.sdk\usr\bin doesn't exist.
The runtime dlls are in S:\Program Files\swift\runtime-development\usr\bin.


compnerd (approved):
Thanks!

PR #567, opened 2023-06-12
Turn off safeseh for foundation on X86.

hjyamauchi:


build.ps1
```diff
@@ -429,6 +429,10 @@ function Build-CMakeProject
     $SwiftArgs.Add("-Xlinker /INCREMENTAL:NO") | Out-Null
     $SwiftArgs.Add("-Xlinker /DEBUG:DWARF") | Out-Null
 
+    # Turn off safeseh for lld as it has safeseh enabled by default
+    # and fails with an ICU data object file icudt69l_dat.obj.
+    $SwiftArgs.Add("-Xlinker /SAFESEH:NO") | Out-Null
```
compnerd, line 434: I don't think that we should disable Safe SEH globally, it has some benefits for attack mitigation.

hjyamauchi: If it was disabled before the lld switch, not much was lost? Is there a way to disable it selectively?

hjyamauchi: Maybe I misunderstood "globally". We could disable safe seh in the `Build-Foundation` case only.

compnerd: Yeah, I think that we should be able to do that.  We can add `-D CMAKE_SHARED_LINKER_FLAGS="/SAFESEH:NO"` to `Build-Foundation`.

hjyamauchi: I saw that `CMAKE_SHARED_LINKER_FLAGS` had no effect on the swift linking. Updated it so it passes the option through a parameter instead.

build.ps1
```diff
@@ -357,7 +357,8 @@ function Build-CMakeProject
     [string[]] $UseBuiltCompilers = @(), # ASM,C,CXX,Swift
     [string] $SwiftSDK = "",
     [hashtable] $Defines = @{},
-    [string[]] $BuildTargets = @()
+    [string[]] $BuildTargets = @(),
+    [string] $ExtraSwiftArgs = ""
```
compnerd, line 361: This is unnecessary

build.ps1
```diff
@@ -433,6 +434,8 @@ function Build-CMakeProject
     $SwiftArgs.Add("-Xlinker /OPT:REF") | Out-Null
     $SwiftArgs.Add("-Xlinker /OPT:ICF") | Out-Null
 
+    $SwiftArgs.Add($ExtraSwiftArgs) | Out-Null
```
compnerd, line 437: This is unnecessary

build.ps1
```diff
@@ -823,6 +826,11 @@ function Build-Foundation($Arch, [switch]$Test = $false)
       $Targets = @("default", "install")
     }
 
+    # Turn off safeseh for lld as it has safeseh enabled by default
+    # and fails with an ICU data object file icudt69l_dat.obj. This
+    # matters to X86 only.
+    $ExtraSwiftArgs = "-Xlinker /SAFESEH:NO"
```
compnerd, line 832: We should conditionalise it to x86 and pass it as `CMAKE_Swift_FLAGS`.

compnerd (changes requested):

build.ps1
hjyamauchi: Done

build.ps1
```diff
@@ -823,6 +823,15 @@ function Build-Foundation($Arch, [switch]$Test = $false)
       $Targets = @("default", "install")
     }
 
+    $SwiftFlags = ""
+    if ($Arch -eq $ArchX86)
+    {
```
compnerd, line 828:
```suggestion
    if ($Arch -eq $ArchX86) {
```

Everything is now uniformly K&R.

compnerd (approved):

build.ps1
```diff
@@ -823,6 +823,14 @@ function Build-Foundation($Arch, [switch]$Test = $false)
       $Targets = @("default", "install")
     }
 
+    $SwiftFlags = ""
+    if ($Arch -eq $ArchX86) {
+      # Turn off safeseh for lld as it has safeseh enabled by default
+      # and fails with an ICU data object file icudt69l_dat.obj. This
+      # matters to X86 only.
+      $SwiftFlags = "-Xlinker /SAFESEH:NO"
+    }
```
compnerd, line 832:
```suggestion
    # Turn off safeseh for lld as it has safeseh enabled by default
    # and fails with an ICU data object file icudt69l_dat.obj. This
    # matters to X86 only.
    $SwiftFlags = if ($Arch -eq $ArchX86) { "-Xlinker /SAFESEH:NO" } else { "" }
```

build.ps1
```diff
@@ -834,6 +839,7 @@ function Build-Foundation($Arch, [switch]$Test = $false)
         CMAKE_INSTALL_PREFIX = "$($Arch.SDKInstallRoot)\usr";
         CMAKE_SYSTEM_NAME = "Windows";
         CMAKE_SYSTEM_PROCESSOR = $Arch.CMakeName;
+        CMAKE_Swift_FLAGS = $SwiftFlags;
```
compnerd, line 842:
```suggestion
        # Turn off safeseh for lld as it has safeseh enabled by default
        # and fails with an ICU data object file icudt69l_dat.obj. This
        # matters to X86 only.
        CMAKE_Swift_FLAGS = if ($Arch -eq $ArchX86) { "-Xlinker /SAFESEH:NO" } else { "" };
```

build.ps1
```diff
@@ -823,6 +823,11 @@ function Build-Foundation($Arch, [switch]$Test = $false)
       $Targets = @("default", "install")
     }
 
+    # Turn off safeseh for lld as it has safeseh enabled by default
+    # and fails with an ICU data object file icudt69l_dat.obj. This
+    # matters to X86 only.
+    $SwiftFlags = if ($Arch -eq $ArchX86) { "-Xlinker /SAFESEH:NO" } else { "" }
+
```
compnerd, line 830:
```suggestion
```

PR #571, opened 2023-06-14
build: disable debug information on release builds

compnerd:
Disable the default build from emitting debug information.  This will speed up the build and reduce the output size.  This is particularly important for the Swift side which will emit a large amount of metadata into the final binaries which is not stripped.

This is intended to be the default build type for CI.  For local builds, developers may use `-BuildType RelWithDebInfo` to get debug information emitted.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -352,8 +352,11 @@ function Build-CMakeProject {
   TryAdd-KeyValue $Defines CMAKE_BUILD_TYPE $BuildType
   TryAdd-KeyValue $Defines CMAKE_MT "mt"
 
-  $CFlags = "/GS- /Gw /Gy /Oi /Oy /Zi /Zc:inline"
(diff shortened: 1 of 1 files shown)


build.ps1
```diff
@@ -352,8 +352,11 @@ function Build-CMakeProject {
   TryAdd-KeyValue $Defines CMAKE_BUILD_TYPE $BuildType
   TryAdd-KeyValue $Defines CMAKE_MT "mt"
 
-  $CFlags = "/GS- /Gw /Gy /Oi /Oy /Zi /Zc:inline"
-  $CXXFlags = "/GS- /Gw /Gy /Oi /Oy /Zi /Zc:inline /Zc:__cplusplus"
+  $IsRelease = $Defines["CMAKE_BUILD_TYPE"] == "Release"
```
tristanlabelle, line 355: `-eq`

tristanlabelle (approved):

PR #573, opened 2023-06-14
Support specifying the debug info format for C/C++ and Swift code

tristanlabelle:
Dwarf is better for Swift but not supported by WinDBG and Windows tooling, so depending on scenarios we may want different combinations of debugging format.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -28,6 +28,12 @@ similar to the /Library or ~/Library directories on macOS.
 .PARAMETER BuildType
 The CMake build type to use, one of: Release, RelWithDebInfo, Debug.
 
+.PARAMETER CDebugFormat
+The debug information format for C/C++ code: dwarf or codeview.
+
+.PARAMETER SwiftDebugFormat
+The debug information format for Swift code: dwarf or codeview.
+
 .PARAMETER SDKs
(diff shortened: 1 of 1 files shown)


build.ps1
```diff
@@ -404,18 +412,16 @@ function Build-CMakeProject {
...
 
     # Debug Information
-    if ($IsRelease) {
-      $SwiftArgs.Add("-gnone") | Out-Null
+    if ($GenerateDebugInfo) {
+      if ($SwiftDebugFormat -eq "dwarf") {
+        $SwiftArgs.Add("-g -Xlinker /DEBUG:DWARF -use-ld=lld-link") | Out-Null
+      } else {
+        $SwiftArgs.Add("-g -debug-info-format=codeview -Xlinker -debug") | Out-Null
+      }
```
compnerd, line 420: One thing that was nice about the slightly more repetitive variant is that we could easily see the difference between the compile and link flags.  We do loose that here though we gain some brevity.

compnerd (approved):

build.ps1
tristanlabelle: I'm not super opinionated here but I did get annoyed by all those `| Out-Null`'s.

PR #575, opened 2023-06-16
build: add `swift-inspect` to the build

compnerd:
This is part of the default toolchain distribution and needs to be built for packaging as well as development.  Add this to the build set.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -1322,6 +1322,16 @@ function Install-HostToolchain() {
   Copy-Item -Force $BinaryCache\7\bin\swift-driver.exe $ToolchainInstallRoot\usr\bin\swiftc.exe
 }
 
+function Build-Inspect() {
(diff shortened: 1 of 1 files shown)


build.ps1
```diff
@@ -1322,6 +1322,16 @@ function Install-HostToolchain() {
   Copy-Item -Force $BinaryCache\7\bin\swift-driver.exe $ToolchainInstallRoot\usr\bin\swiftc.exe
 }
 
+function Build-Inspect() {
+  $OutDir = Join-Path -Path $HostArch.BinaryRoot -ChildPath swift-inspect
```
tristanlabelle, line 1326: Fancier than everywhere else's `"$($HostArch.BinaryRoot)\swift-inspect"`!

tristanlabelle (approved):

build.ps1
compnerd: We should clean that up eventually and use `Join-Path`.

hjyamauchi:
lgtm

PR #576, opened 2023-06-16
build: add `swift-format` to the build

compnerd:
`swift-format` is part of the default toolchain distribution and must be built to be packaged.


tristanlabelle (approved):

build.ps1
```diff
@@ -1432,6 +1445,7 @@ Install-HostToolchain
 
 if (-not $SkipBuild) {
   Build-Inspect $HostArch
+  Build-Format $HostArch
```
hjyamauchi, line 1448: It might be worth considering running the swift-format tests in the future in this script as in the toolchain yml.

PR #580, opened 2023-06-23
build: add a workaround for CMake 3.26.0

compnerd:
MSVC distributes CMake 3.26.0, which does not properly handle `clang-cl`'s output for `/showIncludes`.  Add a workaround until MSVC updates the CMake distributed to 3.26.3 or newer.

See: ninja-build/ninja#2280

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -381,6 +381,9 @@ function Build-CMakeProject {
   if ($UseBuiltCompilers.Contains("C")) {
     TryAdd-KeyValue $Defines CMAKE_C_COMPILER "$BinaryCache\1\bin\clang-cl.exe"
     TryAdd-KeyValue $Defines CMAKE_C_COMPILER_TARGET $Arch.LLVMTarget
+
+    TryAdd-KeyValue $Defines CMAKE_CL_SHOWINCLUDES_PREFIX "Note: including file: "
+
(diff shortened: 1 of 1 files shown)


build.ps1
```diff
@@ -381,6 +381,9 @@ function Build-CMakeProject {
   if ($UseBuiltCompilers.Contains("C")) {
     TryAdd-KeyValue $Defines CMAKE_C_COMPILER "$BinaryCache\1\bin\clang-cl.exe"
     TryAdd-KeyValue $Defines CMAKE_C_COMPILER_TARGET $Arch.LLVMTarget
+
+    TryAdd-KeyValue $Defines CMAKE_CL_SHOWINCLUDES_PREFIX "Note: including file: "
```
tristanlabelle, line 385:
```suggestion
    # Workaround for a CMake 3.26.0 bug with `/showIncludes` (fixed in CMake 3.26.3)
    TryAdd-KeyValue $Defines CMAKE_CL_SHOWINCLUDES_PREFIX "Note: including file: "
```

tristanlabelle (approved):

hjyamauchi (commented):
Nice. With this we won't need the local workaround anymore, I take.

compnerd:
Correct.  I'd been carrying this change, but now that we have a better understanding we can add the workaround.  If we can gate it on the version, that would be even better.

PR #581, opened 2023-06-23
build: repair the ARM64 build of ICU on ARM64

compnerd:
When building on ARM64 ensure that we build the tools as there is no need to have the X64 variant be used for this.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -759,7 +759,7 @@ function Build-ICU($Arch) {
     }
   }
 
-  if ($Arch -eq $ArchARM64) {
+  if ($Arch -eq $ArchARM64 -and $HostArch -ne $ArchARM64) {
     # Use previously built x64 tools
     $BuildToolsDefines = @{
       BUILD_TOOLS = "NO";


build.ps1
```diff
@@ -759,7 +759,7 @@ function Build-ICU($Arch) {
     }
   }
 
-  if ($Arch -eq $ArchARM64) {
+  if ($Arch -eq $ArchARM64 -and $HostArch -ne $ArchARM64) {
```
tristanlabelle, line 762: Since the body of the if explicitly mentions `$ArchX64`, it should probably appear in the condition? `$HostArch -eq $ArchX64`?

tristanlabelle (approved):

hjyamauchi (approved):
Side question: is this ARM64 build referring to a cross-build on X64 host?

compnerd:
The fix is for the native compilation (ARM64 -> ARM64), but the condition is for the cross-compilation case as that requires that we build the ICU tools for the build.

PR #583, opened 2023-06-26
Add CMake version check for /showIncludes workaround

tristanlabelle:
To avoid hardcoding `CMAKE_CL_SHOWINCLUDES_PREFIX` unless we need it since it can break builds on non-English machines.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -330,7 +330,7 @@ function Append-FlagsDefine([hashtable]$Defines, [string]$Name, [string]$Value) 
   }
 }
 
-function Test-CMakeMinVersion([int]$Major, [int]$Minor, [int]$Patch = 0) {
+function Test-CMakeAtLeast([int]$Major, [int]$Minor, [int]$Patch = 0) {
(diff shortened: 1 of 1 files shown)


build.ps1
```diff
@@ -330,6 +330,18 @@ function Append-FlagsDefine([hashtable]$Defines, [string]$Name, [string]$Value)
   }
 }
 
+function Test-CMakeMinVersion([int]$Major, [int]$Minor, [int]$Patch = 0) {
```
compnerd, line 333: Does `Test-CMakeAtLeast` fit the naming scheme for PS?  If so, that seems like a nicer name.

compnerd (approved):
Thanks for improving this!  I wanted this, but didn't want to futz with PS.

build.ps1
tristanlabelle: I don't know if it's legit but I took your suggestion.

hjyamauchi:
Nice!

PR #584, opened 2023-06-28
build.ps1: Make cmake invocation explicit for version testing

tristanlabelle:
`&` tells powershell that we're executing a command and not looking for a function.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -331,7 +331,7 @@ function Append-FlagsDefine([hashtable]$Defines, [string]$Name, [string]$Value) 
 }
 
 function Test-CMakeAtLeast([int]$Major, [int]$Minor, [int]$Patch = 0) {
-  $CMakeVersionString = @(cmake --version)[0]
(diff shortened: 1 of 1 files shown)


build.ps1
```diff
@@ -331,7 +331,7 @@ function Append-FlagsDefine([hashtable]$Defines, [string]$Name, [string]$Value)
 }
 
 function Test-CMakeAtLeast([int]$Major, [int]$Minor, [int]$Patch = 0) {
-  $CMakeVersionString = @(cmake --version)[0]
+  $CMakeVersionString = @(& cmake.exe --version)[0]
```
compnerd, line 334: We also have `Invoke-Command` IIRC, but I suppose that we don't want to log this.

compnerd (approved):

build.ps1
tristanlabelle: Yeah, that was my thinking

PR #585, opened 2023-06-28
build.ps1: Fix cmake.exe version checks when not in %Path%.

tristanlabelle:
We shouldn't try to test the `cmake.exe` version without being in a developer command shell because we may be using Visual Studio's `cmake.exe`, which gets added to the path by `Invoke-VsDevShell`.


hjyamauchi (approved):

compnerd (approved):
adding `?w=1` (or hiding whitespace changes) helps make the diff more readable.

PR #590, opened 2023-07-13
GHA: centralise computation of tag name

compnerd:
We already compute the tag name in the context build, reuse the memoised value rather than re-computing it.  This avoids having the same logic littered across the workflow which is the motivating factor behind the change.


tristanlabelle (approved):
Nice! We could have built across a day boundary and ended up with two different dates 😱

compnerd:
@tristanlabelle yeah ... and this is why DevOps is a thing.  Clearly computers were a mistake.

PR #591, opened 2023-07-18
Remove SwiftFormat from build

tristanlabelle:
We don't need this anymore as the [SwiftFormat repository](https://github.com/nicklockwood/SwiftFormat) is now building this and [producing releases](https://github.com/nicklockwood/SwiftFormat/releases).


compnerd (approved):
Who exactly is "we" in the commit message?

PR #596, opened 2023-07-31
azure: look for VS BuildTools as well

lxbndr:
A bit of quality of life improvement. I have BuildTools installed, not Visual Studio. `vswhere` ignores BuildTools by default😔. This is one of the changes I have to make every time I try to run pipelines on a self-hosted runner. Would be nice to have it in the base🙏


compnerd (approved):
Thanks!  I guess I forgot to do that change here and only did it in the script.

PR #597, opened 2023-08-01
Add swift-format stage

lxbndr:
This attempt uses `build.ps1` as the tool to restructure SDK/Toolchain. While we probably want to avoid using `build.ps1` this way, this still works and proves the concept. I would continue working on standalone restructuring step.

Also, I didn't manage to run the whole pipeline on Azure-hosted agent (i.e. just like original pipeline), because it exceeded 6h quota again😔But testing partially on self-hosted runner demonstrated promising results.

### DIFF
--- a/.azure/vs2022.yml
+++ b/.azure/vs2022.yml
@@ -57,6 +57,11 @@ resources:
       name: apple/swift-cmark
       ref: refs/heads/gfm
       type: github
+    - repository: apple/swift-format
+      endpoint: GitHub
+      name: apple/swift-format
+      ref: refs/heads/main
+      type: github
     - repository: apple/swift-corelibs-libdispatch
       endpoint: GitHub
       name: apple/swift-corelibs-libdispatch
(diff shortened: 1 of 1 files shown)


.azure/vs2022.yml
```diff
@@ -1540,9 +1545,61 @@ stages:
...
+              targetType: 'inline'
+              script: |
+                # Prepare input data for build.ps1
+                New-Item -ItemType Directory -ErrorAction Ignore $(Agent.BuildDirectory)/$(platform) | Out-Null
+                New-Item -ItemType Directory -ErrorAction Ignore $(Agent.BuildDirectory)/7/bin | Out-Null
+                Move-Item $(Agent.BuildDirectory)/Library/Developer/Platforms/Windows.platform $(Agent.BuildDirectory)/$(platform)/
+                Copy-Item -Force $(Agent.BuildDirectory)/Library/Developer/Toolchains/unknown-Asserts-development.xctoolchain/usr/bin/swift-driver.exe $(Agent.BuildDirectory)/7/bin/
+
+                # Ask build.ps1 to create a correct toolchain/sdk layout
+                $(Build.SourcesDirectory)/swift-build/build.ps1 -SkipRedistInstall -SkipBuild -SkipPackaging -SDKs X64 -BinaryCache $(Agent.BuildDirectory) -LibraryRoot $(Agent.BuildDirectory)/Library
```
compnerd, line 1589: One thing to be aware of - this is going to go away.  The desire is to move this script into apple/swift.

.azure/vs2022.yml
```diff
@@ -1540,9 +1545,61 @@ stages:
...
+            fetchDepth: 1
+          - checkout: apple/swift-format
+            fetchDepth: 1
+          - task: PowerShell@2
+            inputs:
+              targetType: 'inline'
+              script: |
+                # Prepare input data for build.ps1
+                New-Item -ItemType Directory -ErrorAction Ignore $(Agent.BuildDirectory)/$(platform) | Out-Null
+                New-Item -ItemType Directory -ErrorAction Ignore $(Agent.BuildDirectory)/7/bin | Out-Null
```
compnerd, line 1584: This also is a bit of a discomfort - the numbers have been renumbered in the past.  We shouldn't be relying on them to be consistent across builds.

lxbndr:
@compnerd I made it standalone. Based on the code from `build.ps1`. Naming and code style comments are welcome too. I am not sure I got the spirit from the existing code🤔

.azure/vs2022.yml
```diff
@@ -1540,9 +1545,121 @@ stages:
...
+                  New-Item -ItemType Directory -ErrorAction Ignore $Dst | Out-Null
+                  Copy-Item -Force -Recurse $Src $Dst
+                }
+
+                Remove-Item $LibraryRoot -Recurse -ErrorAction Ignore
+                
+                # Install toolchain
+                #-----------------------------------------------------------------------------
+                New-Item -ItemType Directory -ErrorAction Ignore $LibraryRoot | Out-Null
+                Move-Item $ToolchainBinaryCache/Library/Developer $LibraryRoot
```
compnerd, line 1603: Can we use Copy-Directory instead?

.azure/vs2022.yml
```diff
@@ -1540,9 +1545,121 @@ stages:
...
+
+                function Copy-File($Src, $Dst) {
+                  $DstDir = [IO.Path]::GetDirectoryName($Dst)
+                  New-Item -ItemType Directory -ErrorAction Ignore $DstDir | Out-Null
+                  Copy-Item -Force $Src $Dst
+                }
+
+                function Copy-Directory($Src, $Dst) {
+                  New-Item -ItemType Directory -ErrorAction Ignore $Dst | Out-Null
+                  Copy-Item -Force -Recurse $Src $Dst
```
compnerd, line 1595: I think that we want to move instead - we are space constrained on the hosted runners.

.azure/vs2022.yml
```diff
@@ -1540,9 +1545,121 @@ stages:
...
+                $DevtoolsBinaryCache = "$(Agent.BuildDirectory)/devtools-amd64"
+                $PlatformBinaryCache = "$(Agent.BuildDirectory)/windows-sdk-$(arch)"
+
+                $SDKSourceRoot = "$PlatformBinaryCache/Library/Developer/Platforms/Windows.platform/Developer/SDKs/Windows.sdk"
+                $XCTestSourceRoot = "$PlatformBinaryCache/Library/Developer/Platforms/Windows.platform/Developer/Library/XCTest-development"
+
+                function Copy-File($Src, $Dst) {
+                  $DstDir = [IO.Path]::GetDirectoryName($Dst)
+                  New-Item -ItemType Directory -ErrorAction Ignore $DstDir | Out-Null
+                  Copy-Item -Force $Src $Dst
```
compnerd, line 1590: I think that we should use move as we are space constrained on hosted runners

.azure/vs2022.yml
```diff
@@ -1540,9 +1545,121 @@ stages:
...
+                Copy-File $XCTestSourceRoot/usr/lib/swift/windows/XCTest.lib $XCTestInstallRoot/usr/lib/swift/windows/$(platform)/
+                Copy-File $XCTestSourceRoot/usr/lib/swift/windows/$(platform)/XCTest.swiftmodule $XCTestInstallRoot/usr/lib/swift/windows/XCTest.swiftmodule/$(platform).swiftmodule
+                Copy-File $XCTestSourceRoot/usr/lib/swift/windows/$(platform)/XCTest.swiftdoc $XCTestInstallRoot/usr/lib/swift/windows/XCTest.swiftmodule/$(platform).swiftdoc
+
+                Copy-File $PlatformBinaryCache/Library/Developer/Platforms/Windows.platform/Info.plist $PlatformInstallRoot/
+                Copy-File $SDKSourceRoot/SDKSettings.plist $SDKInstallRoot/
+
+                # Set up environment
+                #-----------------------------------------------------------------------------
+                Write-Host "##vso[task.setvariable variable=PATH;]${env:Path};$(Agent.BuildDirectory)\windows-sdk-$(arch)\Library\Developer\Platforms\Windows.platform\Developer\SDKs\Windows.sdk\usr\bin;$(Agent.BuildDirectory)\Library\Developer\Toolchains\unknown-Asserts-development.xctoolchain\usr\bin"
```
compnerd, line 1652: This is probably fine but I would’ve prepended rather than appended.

.azure/vs2022.yml
lxbndr: Sure, but I think I'll rename it to `Move-Directory`, as the requests below requiring to replace `Copy-Item` with `Move-Item`.

.azure/vs2022.yml
lxbndr: Ok. And it is also faster than copying. Speed for free!

lxbndr:
I forgot that moving folders in Powershell is not like copying. We can't just merge two folder structures with Move-Item. Had to improvise, but it seems to work in the end. Now everything is moved, not copied. Also, the PATH is prepended.

compnerd (approved):
This looks good to me!  Might be a good idea to test it ...

compnerd:
> /.azure/vs2022.yml: (Line: 1656, Col: 15, Idx: 85905) - (Line: 1656, Col: 15, Idx: 85905): While scanning a simple key, could not find expected ':'.

lxbndr:
Oops, my apologies. I was shuffling code back and forward when testing this, and there is always a place for a mistake. But I had to check for validity in first place before pushing this😔 Now it should be ok.

PR #600, opened 2023-08-03
build: deploy CxxShim module on Windows

compnerd:
The installer will copy this module as well, but we failed to deploy it from the staging to the local installation which prevents the local deployment from being used for testing C++ interop.


hjyamauchi (commented):
lgtm

PR #601, opened 2023-08-06
Azure: use vfs overlay in devtools

lxbndr:


.azure/vs2022.yml
```diff
@@ -1062,6 +1062,8 @@ stages:
                   # runtime.
                   plistlib.dump({ 'DefaultProperties': { 'DEFAULT_USE_RUNTIME': 'MD' } }, plist)
               failOnStderr: true
+          - script: |
+              copy $(Agent.BuildDirectory)\swift\stdlib\windows-vfs-overlay.yaml $(Build.StagingDirectory)\Library\Developer\Platforms\Windows.platform\Developer\SDKs\Windows.sdk\usr\share
```
compnerd, line 1066: Mind using `Copy-Item`?

compnerd (approved):
I'd like to try to get a run through on Azure.  I am concerned that we might need some more changes to the packaging step (from other changes that have gone in to the installer).

.azure/vs2022.yml
lxbndr: @compnerd Done

PR #603, opened 2023-08-08
Upload MSM to release artifacts during swift-toolchain build

fischman-bcny:
Example invocation: https://github.com/thebrowsercompany/swift-build/actions/runs/5798835944
Artifacts: TODO

PR #604, opened 2023-08-09
Upload runtime MSM to Release artifacts

fischman-bcny:
Follows up #603 by downloading the MSM from the swift-toolchain.yml action run and uploading it to the GitHub release's artifacts.
Test run: https://github.com/thebrowsercompany/swift-build/actions/runs/5812190767
Result: https://github.com/thebrowsercompany/swift-build/releases/tag/20230809.001-ami contains runtime.amd64.msm
image

### DIFF
--- a/.github/workflows/swift-toolchain.yml
+++ b/.github/workflows/swift-toolchain.yml
@@ -1985,6 +1985,11 @@ jobs:
           name: installer-amd64
           path: ${{ github.workspace }}/tmp
 
+      - uses: actions/download-artifact@v3
+        with:
+          name: runtime.amd64.msm
+          path: ${{ github.workspace }}/tmp
+
       - uses: actions/create-release@v1
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/swift-toolchain.yml
```diff
@@ -1985,6 +1985,11 @@ jobs:
           name: installer-amd64
           path: ${{ github.workspace }}/tmp
 
+      - uses: actions/download-artifact@v3
+        with:
+          name: runtime.amd64.msm
+          path: ${{ github.workspace }}/tmp
```
compnerd, line 1991: Could we pick up all 3 MSMs?  (arm64, x86, and arm64)

fischman-bcny: Sure, but I'd rather get this platform working end-to-end first.
(also, e.g., note we only upload installer-amd64.exe to the Release today)

compnerd: Right, we only upload installer-amd64, but it contains the SDK for all the platforms :)

compnerd: @fischman-bcny pointed out offline that we don't actually package the ARM64 SDK, so ...

fischman-bcny: Per tuple, we're [not packaging the arm64](https://github.com/thebrowsercompany/swift-build/blob/b640eafa68af5eae455ad4ff13dd9298f7ab5f1b/.github/workflows/swift-toolchain.yml#L1763-L1764) SDK yet, so doesn't seem right to pick the MSM yet.
(likely both to come)

PR #605, opened 2023-08-10
build: collect binlogs on the build

compnerd:
This can be helpful to diagnose build failures.  Always collect a binlog so that we can analyse the build process later.  Take the opportunity to clean out an unreferenced variable.  One side effect is that the summary view is enabled with the binarylogger, so disable that explicitly.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -566,7 +566,6 @@ function Build-WiXProject() {
     [hashtable]$Properties = @{}
   )
 
-  $Name = $FileName.Split('.')[0]
   $ArchName = $Arch.VSName
 
   $ProductVersionArg = $ProductVersion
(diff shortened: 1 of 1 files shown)


fischman-bcny (approved):
IDK what any of this is, but if it does what you want, vaya con dios.
PR description might be aided by an example output of this and how one views/uses the resulting binlog.

compnerd:
https://www.msbuildlog.com/ is the tool for viewing the log.  I don't think that sharing the output is helpful (its a binary file).  However, it should be available in `$BinaryCache\$HostArch\msi` as `$Arch-$MSI.binlog`, e.g.

amd64-bld.binlog
amd64-cli.binlog
amd64-dbg.binlog
amd64-ide.binlog
amd64-rtl.binlog
amd64-sdk.binlog
...

hjyamauchi (commented):
lgtm

PR #608, opened 2023-08-10
build: attempt to reuse checkouts for building auxtools

compnerd:
Avoid re-cloning the dependencies for swift-format and swift-docc and reuse the existing checkout using the CI paths.


fischman-bcny (commented):
LGTM assuming it works, in which case IMO add link to passing run in PR description.
Bonus points for before/after links showing the speedup.

fischman-bcny (approved):

hjyamauchi (approved):

compnerd:
build.ps1 is for local builds only, so no CI yet.  @tristanlabelle should hopefully get that fixed soon!

PR #609, opened 2023-08-11
Update VS as workaround to arm64 build slowness

tristanlabelle:
Workaround for MSVC large switch jump table computation taking ~1h30, fixed in 17.7.

### DIFF
--- a/.github/workflows/swift-toolchain.yml
+++ b/.github/workflows/swift-toolchain.yml
@@ -481,6 +481,12 @@ jobs:
           echo "PYTHON_LOCATION_amd64=$env:pythonLocation" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
           echo "PYTHON_LOCATION_arm64=${{ github.workspace }}\pythonarm64.${{ env.PYTHON_VERSION }}\tools" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
 
+      # Workaround for MSVC large switch jump table computation taking ~1h30
+      - name: Update Visual Studio to 17.7
+        if: matrix.arch == 'arm64'
+        run: |
+          & "C:\Program Files (x86)\Microsoft Visual Studio\Installer\vs_installer.exe" update --productId Microsoft.VisualStudio.Product.Community --channelId VisualStudio.17.Release --updateVersion 17.7 --quiet --noRestart
+
       - name: Install llvm
         if: matrix.arch == 'arm64'
         run: |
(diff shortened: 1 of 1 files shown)


compnerd:
It may be resolved (at least @lxbndr had some luck with it), but it is still uncertain!

.github/workflows/swift-toolchain.yml
```diff
@@ -481,6 +481,12 @@ jobs:
           echo "PYTHON_LOCATION_amd64=$env:pythonLocation" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
           echo "PYTHON_LOCATION_arm64=${{ github.workspace }}\pythonarm64.${{ env.PYTHON_VERSION }}\tools" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
 
+      # Workaround for MSVC large switch jump table computation taking ~1h30
+      - name: Update Visual Studio to 17.7
+        if: matrix.arch == 'arm64'
+        run: |
+          & "C:\Program Files (x86)\Microsoft Visual Studio\Installer\vs_installer.exe" update --productId Microsoft.VisualStudio.Product.Community --channelId VisualStudio.17.Release --updateVersion 17.7 --quiet --noRestart
```
compnerd, line 488: I wonder if we can get away with gha-setup-vsdevenv?  IIRC it has some provisions for managing the components.  We should also verify if the builder uses Community or some other SKU.  We do not want to just install another SKU as we are limited on space.

tristanlabelle: Good call! It's enterprise based on: https://github.com/actions/runner-images/blob/main/images/win/Windows2022-Readme.md

tristanlabelle: I also verified locally that getting the wrong SKU won't install an additional instance (because the operation is `update`):

```
[5fec:0014][2023-08-11T16:53:50] Error 0x80131509: [5fec:0014][2023-08-11T16:53:50] Failed to initialize the app in AppInitializerService.InitializeChannelsAsync: System.InvalidOperationException: No instance found with productId Microsoft.VisualStudio.Product.Enterprise and channelId VisualStudio.17.Release

tristanlabelle: From what I can see, `gha-setup-vsdevenv` only has filters to find installations having certain components, it won't trigger updates itself

compnerd:
Sadly don't think that this actually fixes the issue.  I think that we might have to do a different workaround - reducing some optimizations on `StandardLibrary.cpp`.

tristanlabelle:
Too bad :(

PR #612, opened 2023-08-14
Azure: patch llvm config as temporary workaround for slow arm64 builds

lxbndr:
Here we're disabling all optimizations with `/Od`, and then enabling `/Gw /Oi /Oy /Gy /Ob2 /Ot /GF`, which is the same flag set as `/O2` enables, but without "Global Optimizations" (`/Og`).


lxbndr:
@compnerd could you please start the check on your side already? I didn't finish a complete run with this, but a single arm64 build finishes in a reasonable time.

lxbndr:
@compnerd we need a new run for this. I am expecting it to have more chances to complete. Maybe not on first, but on the second/third attempt 😅

compnerd (approved):
Managed to get a build with this!

PR #616, opened 2023-08-16
Support configuring runners through repo variables.

tristanlabelle:
So forks of this repo can use different-sized runner machines.


compnerd:
Can you hoist this into a variable in the top level to avoid the `||` everywhere?

tristanlabelle:
> Can you hoist this into a variable in the top level to avoid the `||` everywhere?

I don't think I can:

image

Only `vars` and `inputs` are candidates and `inputs` wouldn't work because https://github.com/orgs/community/discussions/25218

tristanlabelle:
Hmm maybe it works through the "needs" context

compnerd (approved):

PR #617, opened 2023-08-17
Revert to "XCTest-development" naming

tristanlabelle:
We do want `XCTest-$ProductVersion` at some point, but this must be piped through the installer and `build-toolchain-windows.bat` (if it still exists at that point)


compnerd (changes requested):
There is the change to the plist that needs to done in this partial revert.

compnerd (approved):

PR #619, opened 2023-08-17
Detangle "Install" vs "Image" in variable names

tristanlabelle:
"Image" now always refers to the place where we build the "Library" and "Program Files" that mimic a filesystem image, whereas "Install" refers to places we use as a CMake install destination.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -176,7 +176,7 @@ $HostArch = switch ($NativeProcessorArchName) {
   default { throw "Unsupported processor architecture" }
 }
 
-function Get-InstallDir($Arch) {
+function Get-ImageSwiftRoot($Arch) {
   if ($Arch -eq $HostArch) {
     $ProgramFilesName = "Program Files"
   } elseif ($Arch -eq $ArchX86) {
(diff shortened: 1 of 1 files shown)


build.ps1
```diff
@@ -176,7 +176,7 @@ $HostArch = switch ($NativeProcessorArchName) {
   default { throw "Unsupported processor architecture" }
 }
 
-function Get-InstallDir($Arch) {
+function Get-ImageSwiftRoot($Arch) {
```
compnerd, line 179: What about `Get-ImageRoot` or `Get-SwiftImageRoot`?

build.ps1
```diff
@@ -192,15 +192,16 @@ function Get-InstallDir($Arch) {
...
-$ToolchainInstallRoot = "$(Get-InstallDir $HostArch)\Toolchains\$ProductVersion+Asserts"
-$PlatformInstallRoot = "$(Get-InstallDir $HostArch)\Platforms\Windows.platform"
-$RuntimeInstallRoot = "$(Get-InstallDir $HostArch)\Runtimes\$ProductVersion"
-$SDKInstallRoot = "$PlatformInstallRoot\Developer\SDKs\Windows.sdk"
+$ImageLibraryRoot = "$ImageRoot\Library"
+$ImageSwiftRoot = Get-ImageSwiftRoot $HostArch
+$ImageToolchainRoot = "$ImageSwiftRoot\Toolchains\$ProductVersion+Asserts"
+$ImagePlatformRoot = "$ImageSwiftRoot\Platforms\Windows.platform"
+$ImageRuntimeRoot = "$ImageSwiftRoot\Runtimes\$ProductVersion"
+$ImageSDKRoot = "$ImagePlatformRoot\Developer\SDKs\Windows.sdk"
```
compnerd, line 200: Why the renaming?  The behaviour of `BinaryRoot` should be similar to that of `DESTDIR` in autotools.

compnerd (changes requested):

build.ps1
tristanlabelle: I don't understand? The goal was to not name "Install" something that is not a CMake "Install" destination. Since we have "ImageRoot" as the parameter name, it follows that all of these would be called "Image" and that this distinguishes them from arch-specific "Arch.Install" directories.

build.ps1
tristanlabelle: `Get-ImageRoot` would seem confusing given `$ImageRoot`. I prefixed everything with "Image" for uniformity. Would you prefer "SwiftImageRoot", "ToolchainImageRoot", "PlatformImageRoot", "RuntimeImageRoot" and "SDKImageRoot"?

build.ps1
tristanlabelle: I previously understood that you preferred to keep the "Install" naming to only refer to cmake uses of the term.

build.ps1
compnerd: Yeah, I think that the swapped order it better.

build.ps1
compnerd: Yes (though I think that BinaryRoot is still the better term).  Once you swap the order, I think the renaming will make a bit more sense.  It just seemed like you were changing the variables but I missed the change in the name itself.

tristanlabelle:
Closing due to staleness at this point

PR #621, opened 2023-08-17
Use ImageRoot-based paths when building SPM projects

tristanlabelle:
The $ImageRoot-based paths (especially SDK) have had reshuffling applied to match a freshly installed toolchain's layout.

main version of #620


compnerd (approved):
🤷‍♂️ If it works, ship it.

PR #626, opened 2023-08-21
Azure: free space for toolchain install

lxbndr:
This is adventurous a little, but should save us about 1Gb+ of space.
Hope I didn't mess the condition. I want this be enabled only for Azure hosted runners, and there seems no other way to detect them.


compnerd (commented):
Haha, this is certainly adventurous!  We need this on vs2022-5.9.yml as well.

compnerd:
#yolo - tonight, we test in prod 😂 (I really want to get an updated 5.9 snapshot)

compnerd (approved):

PR #629, opened 2023-08-22
Azure: cleanup previous build data for self-hosted agents

lxbndr:
I am trying to reduce the amount of work I have to repeat every time I run the pipeline locally. E.g. I have to disable archs except of only one, because build folders are reused for different matrix branches (x64, arm64, x86) when running on a self-hosted agent, and LLVM/Swift are not happy on such "incremental" reconfigured builds.

This is approach I tested for some time, and it works pretty well. I am using a marker file to store the actual arch we're building. In case it is already present and contains a different arch, I am doing a cleanup of build output folders (still reusing source cache).

Can't say I am thrilled with such bookmarking, but I don't see any simple way to detect arch. Thought about reading CMake caches, but that seems requiring even more code.

Also, there is a condition on cleanup steps, guarding azure-hosted agents from unnecessary work.

### DIFF
--- a/.azure/vs2022.yml
+++ b/.azure/vs2022.yml
@@ -310,6 +310,23 @@ stages:
             fetchDepth: 1
           - checkout: apple/swift-corelibs-libdispatch
             fetchDepth: 1
+          - task: PowerShell@2
+            inputs:
+              targetType: 'inline'
+              script: |
+                # Use a special file to mark last built arch
+                $CookieFilePath = "$(Agent.BuildDirectory)/arch.txt"
+                if (Test-Path -Path $CookieFilePath) {
+                  $LastBuiltArch = (Get-Content -Path $CookieFilePath | Out-String).Trim()
+                }
+                if ($LastBuiltArch -eq "$(arch)") {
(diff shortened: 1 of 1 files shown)


lxbndr:
Converting this to Draft, as it worth to try implement this using [Pipelines workspace parameters](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/phases?view=azure-devops&tabs=yaml#workspace).

lxbndr:
Well, it seems to be impossible to achieve good results using only standard options. I was aiming these points:
- Do cleanup only **when** it is necessary. Build output is only have to be purged when we're building arch different than previous attempt. Same archs are ok to stay, so incremental builds are possible.
- Remove only **what** is necessary. Azure standard options are: everything, sources, outputs. As we're not storing outputs in default location (due to path length issues), we can't rely on outputs cleanup. And removing everything else is unnecessary at all.

As this change significantly increases the complexity of the pipeline definition, I'd like to not merge it. The change would still be [supported in the fork](https://github.com/lxbndr/swift-build/tree/lxbndr/local-cleanup), if anyone need it.

PR #632, opened 2023-08-25
Fix quoting in CMake args defines

tristanlabelle:
If a cmake define is an argument string, we need to distinguish tokens containing spaces from multiple tokens. For example, we must be able to pass the below as two command line arguments to cmake:
`-D` `CMAKE_Swift_FLAGS=-sdk "S:/Program Files/Swift/Platforms/Windows.platform/Developer/SDKs/Windows.sdk" -g -Xlinker /DEBUG:DWARF -use-ld=lld-link -Xlinker /INCREMENTAL:NO -Xlinker /OPT:REF -Xlinker /OPT:ICF`

Quoting becomes tricky in this case. We need to quote compiler flags arguments when they contain spaces to keep them a single token, and because of Windows cmd or cmake rules, the quotes need to be escaped with backslashes. Which means that we can't replace all backslashes into forward slashes wholesale.

The solution is to support two kinds of entries in the `$Defines` hashtable: single strings and arrays of strings. For defines whose value is an array of strings, we'll run logic to do the proper quoting into a final string that will preserve the tokenization when calling the compiler. This fixes issues with `"-sdk $SDKPath"`, which would be interpreted as more than two tokens if `$SDKPath` contains spaces.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -320,9 +320,9 @@ function TryAdd-KeyValue([hashtable]$Hashtable, [string]$Key, [string]$Value) {
   }
 }
 
-function Append-FlagsDefine([hashtable]$Defines, [string]$Name, [string]$Value) {
+function Append-FlagsDefine([hashtable]$Defines, [string]$Name, [string[]]$Value) {
   if ($Defines.Contains($Name)) {
-    $Defines[$name] += " $Value" 
+    $Defines[$name] = @($Defines[$name]) + $Value
   } else {
     $Defines.Add($Name, $Value)
   }
@@ -354,7 +354,7 @@ function Build-CMakeProject {
     [string[]] $UseMSVCCompilers = @(), # C,CXX
     [string[]] $UseBuiltCompilers = @(), # ASM,C,CXX,Swift
     [string] $SwiftSDK = "",
-    [hashtable] $Defines = @{},
+    [hashtable] $Defines = @{}, # Values are either single strings or arrays of flags
     [string[]] $BuildTargets = @()
   )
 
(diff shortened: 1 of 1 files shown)


build.ps1
```diff
@@ -473,15 +473,34 @@ function Build-CMakeProject {
...
+      # The quoting gets tricky to support defines containing compiler flags args,
+      # some of which can contain spaces, for example `-D` `Flags=-flag "C:/Program Files"`
+      # Avoid backslashes since they are going into CMakeCache.txt,
+      # where they are interpreted as escapes.
+      if ($Define.Value -is [string]) {
+        # Single token value, no need to quote spaces, the splat operator does the right thing.
+        $Value = $Define.Value.Replace("\", "/")
       }
-      $cmakeGenerateArgs += @("-D", "$($Define.Key)=$ValueWithForwardSlashes")
+      else {
```
compnerd, line 484: Coddle please

build.ps1
```diff
@@ -473,15 +473,34 @@ function Build-CMakeProject {
...
+          if ($Value.Length -gt 0) {
+            $Value += " "
+          }
+
+          $ArgWithForwardSlashes = $Arg.Replace("\", "/")
+          if ($ArgWithForwardSlashes.Contains(" ")) {
+            # Quote and escape the quote so it makes it through
+            $Value += "\""$ArgWithForwardSlashes\"""
+          }
+          else {
```
compnerd, line 497:
```suggestion
          } else {
```

build.ps1
tristanlabelle: Coddled up

build.ps1
```diff
@@ -473,15 +473,32 @@ function Build-CMakeProject {
...
-        $ValueWithForwardSlashes = $ValueWithForwardSlashes.Replace("<SDK>", "\`"$SwiftSDK\`"")
+      # The quoting gets tricky to support defines containing compiler flags args,
+      # some of which can contain spaces, for example `-D` `Flags=-flag "C:/Program Files"`
+      # Avoid backslashes since they are going into CMakeCache.txt,
+      # where they are interpreted as escapes.
+      if ($Define.Value -is [string]) {
+        # Single token value, no need to quote spaces, the splat operator does the right thing.
+        $Value = $Define.Value.Replace("\", "/")
+      } else {
+        # Flags array, multiple tokens, quoting needed for tokens containing spaces
```
compnerd, line 484: I wish that there was a way to assert that it is a enumerable type (i.e. array).

compnerd (approved):

build.ps1
tristanlabelle: `-is [array]` works

PR #637, opened 2023-08-28
Fix installer stage copy step

tristanlabelle:
My understanding is that Powershell has funny scope resolution rules. A variable read will try the local scope and then the script scope, but a `+=` will modify a local scoped variable, which isn't defined.


tristanlabelle:
Already cherry-picked

PR #638, opened 2023-08-28
build: guard against null args (e.g. default installs)

compnerd:
This corrects an error scenario when building, the call to `.replace` would fail if `$Arg` was `$null`.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -374,7 +374,7 @@ function Build-CMakeProject {
 
     # Add additional defines (unless already present)
     $Defines = $Defines.Clone()
-  
+
     TryAdd-KeyValue $Defines CMAKE_BUILD_TYPE $BuildType
     TryAdd-KeyValue $Defines CMAKE_MT "mt"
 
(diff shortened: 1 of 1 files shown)


build.ps1
```diff
@@ -488,12 +488,14 @@ function Build-CMakeProject {
             $Value += " "
           }
 
-          $ArgWithForwardSlashes = $Arg.Replace("\", "/")
-          if ($ArgWithForwardSlashes.Contains(" ")) {
-            # Quote and escape the quote so it makes it through
-            $Value += "\""$ArgWithForwardSlashes\"""
-          } else {
-            $Value += $ArgWithForwardSlashes
+          if ($Arg -ne $null) {
```
tristanlabelle, line 491: What is inserting `$null` in here? We should rather intercept it in `TryAdd-KeyValue`

compnerd: Unknown, but `$Arg` was unbound, and thus `.Replace` would fail.

tristanlabelle: I found it! gimme a sec

tristanlabelle: https://github.com/compnerd/swift-build/pull/639 . I wasn't hitting it because I build for codeview

PR #642, opened 2023-08-29
build: repair the inclusion specifier

compnerd:
The SDKs are included by the `INCLUDE_[ARCH]_SDK` indicator which was missing the `_SDK` suffix preventing the inclusion.  Correct this to repair the packaging.


tristanlabelle (approved):
Could be worth consolidating the repeated interpolated expression

PR #649, opened 2023-09-01
build.ps1: Add -BuildTo

tristanlabelle:
Allows specifying a build step after which the build script invocation should terminate, for example `-BuildTo ToolsSupportCore`.

Currently only supports build steps (including the installer build), and cannot distinguish between different architectures for SDK/runtime projects.


tristanlabelle:
@compnerd Would this functionality still be useful?

compnerd (approved):

PR #650, opened 2023-09-06
build: adjust the build to enable early syntax parser

compnerd:
The early syntax parser is required to support macros on Windows.  This wires up the pinned toolchain fetch, extract, and wiring to build and link against the early syntax parser.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -42,6 +42,9 @@ An array of architectures for which the Swift SDK should be built.
 The product version to be used when building the installer.
 Supports semantic version strings.
 
+.PARAMETER PinnedToolchain
+The development toolchain snapshot to build the early components with.
+
 .PARAMETER WinSDKVersion
 The version number of the Windows SDK to be used.
 Overrides the value resolved by the Visual Studio command prompt.
@@ -83,6 +86,7 @@ param(
   [string] $SwiftDebugFormat = "dwarf",
   [string[]] $SDKs = @("X64","X86","Arm64"),
   [string] $ProductVersion = "0.0.0",
+  [string] $PinnedToolchain = "swift-DEVELOPMENT-SNAPSHOT-2023-08-12-a",
   [string] $WinSDKVersion = "",
   [switch] $SkipBuild = $false,
   [switch] $SkipRedistInstall = $false,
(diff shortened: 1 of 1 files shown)


build.ps1
```diff
@@ -401,6 +405,28 @@ function Ensure-WindowsSDK {
   }
 }
 
+function Ensure-SwiftToolchain($Arch) {
+  if (Test-Path "$BinaryCache\toolchains\$PinnedToolchain\PFiles64\Swift\Toolchains\0.0.0+Asserts\usr\bin\swiftc.exe" -IsValid) {
```
tristanlabelle, line 409: Is `-IsValid` what we want? It looks like it's only a syntactic check.

image

build.ps1
```diff
@@ -705,6 +731,25 @@ function Build-WiXProject() {
...
+    $env:SDKROOT = "$BinaryCache\toolchains\$PinnedToolchain\PFiles64\Swift\Platforms\Windows.platform\Developer\SDKs\Windows.sdk"
+
+    Build-CMakeProject `
+      -Src $SourceCache\swift-syntax `
+      -Bin $BinaryCache\99 `
+      -InstallTo "$($Arch.ToolchainInstallRoot)\usr" `
+      -Arch $Arch `
+      -BuildTargets default `
+      -Defines @{
+        CMAKE_Swift_COMPILER = "$BinaryCache\toolchains\$PinnedToolchain\PFiles64\Swift\Toolchains\0.0.0+Asserts\usr\bin\swiftc.exe";
```
tristanlabelle, line 746: Do we need to repeat the full path if it's in the Path above?

build.ps1
```diff
@@ -775,6 +822,8 @@ function Build-Compilers($Arch, [switch]$Test = $false) {
         # give us a sligtly faster build.
         CMAKE_BUILD_TYPE = "Release";
         CMAKE_INSTALL_PREFIX = "$($Arch.ToolchainInstallRoot)\usr";
+        CMAKE_Swift_COMPILER = "$BinaryCache\toolchains\$PinnedToolchain\PFiles64\Swift\Toolchains\0.0.0+Asserts\usr\bin\swiftc.exe";
+        CMAKE_Swift_FLAGS = "-sdk $BinaryCache\toolchains\$PinnedToolchain\PFiles64\Swift\Platforms\Windows.platform\Developer\SDKs\Windows.sdk";
```
tristanlabelle, line 826: @("-sdk", "$BinaryCache\toolchains\$PinnedToolchain\PFiles64\Swift\Platforms\Windows.platform\Developer\SDKs\Windows.sdk")

We need this so we can escape spaces properly (ie within one of the $Vars but not between -sdk and the path)

build.ps1
```diff
@@ -83,6 +86,7 @@ param(
   [string] $SwiftDebugFormat = "dwarf",
   [string[]] $SDKs = @("X64","X86","Arm64"),
   [string] $ProductVersion = "0.0.0",
+  [string] $PinnedToolchain = "swift-DEVELOPMENT-SNAPSHOT-2023-08-12-a",
```
tristanlabelle, line 89: maybe `PreviousToolchain`? It describes a bit better why we need it.

tristanlabelle (changes requested):
The Flags part is needs to be an array.

build.ps1
compnerd: It isn't a previous toolchain, its a pinned toolchain that we use to build parts of it.

build.ps1
compnerd: Yes, we should always use absolute paths even if it is in the path.  This is super critical as we don't want accidental usage of the wrong compiler.  (`S:\b\1\bin` will always be given precedence).

build.ps1
compnerd: No, this is wrong.

build.ps1
compnerd: Ugh, good point!

build.ps1
compnerd: In some cases this may even be a future toolchain (like on 5.9).

build.ps1
```diff
@@ -705,6 +738,25 @@ function Build-WiXProject() {
...
+    Build-CMakeProject `
+      -Src $SourceCache\swift-syntax `
+      -Bin $BinaryCache\99 `
+      -InstallTo "$($Arch.ToolchainInstallRoot)\usr" `
+      -Arch $Arch `
+      -BuildTargets default `
+      -Defines @{
+        BUILD_SHARED_LIBS = "YES";
+        CMAKE_Swift_COMPILER = "$BinaryCache\toolchains\$PinnedToolchain\PFiles64\Swift\Toolchains\0.0.0+Asserts\usr\bin\swiftc.exe";
+        CMAKE_Swift_FLAGS = "-sdk ${env:SDKROOT}";
```
tristanlabelle, line 755: Oops, another instance of needing `@("-sdk", $env:SDKROOT)`

tristanlabelle (approved):
Looks good but for the Flags with spaces!

build.ps1
compnerd: I uh, yes, it must've been a clone that did this!

hjyamauchi:
Q: does this early-build toolchain installation conflict with the toolchain that you had installed? Or is this orthogonal?

compnerd:
It shouldn't as we explicitly take steps to download and extract the contents rather than run the installer.  This renders the toolchain inert and you must go out of your way to use it.

PR #654, opened 2023-09-20
Enable building the arm64 installer in the GitHub CI

tristanlabelle:
Validated here: https://github.com/thebrowsercompany/swift-build/actions/runs/6240975701


compnerd (approved):
I'm sure that I had a good reason to typo `triple` as `target` ... right? _right_?

PR #656, opened 2023-09-25
Fix curl error when running build.ps1 after nuking the binary cache.

tristanlabelle:
By default `curl` won't create intermediate path directories. It would work on a rerun because `Expand-Archive` creates the directory before failing due to not finding the zip.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -411,7 +411,7 @@ function Ensure-SwiftToolchain($Arch) {
   }
 
   if (-not (Test-Path $BinaryCache\wix-4.0.1.zip)) {
-    curl.exe -sL https://www.nuget.org/api/v2/package/wix/4.0.1 -o $BinaryCache\wix-4.0.1.zip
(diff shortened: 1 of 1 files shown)


hjyamauchi (approved):

build.ps1
```diff
@@ -411,7 +411,7 @@ function Ensure-SwiftToolchain($Arch) {
   }
 
   if (-not (Test-Path $BinaryCache\wix-4.0.1.zip)) {
-    curl.exe -sL https://www.nuget.org/api/v2/package/wix/4.0.1 -o $BinaryCache\wix-4.0.1.zip
+    Invoke-Program curl.exe -sL https://www.nuget.org/api/v2/package/wix/4.0.1 --output $BinaryCache\wix-4.0.1.zip --create-dirs
```
compnerd, line 414: 😂 you and your long arguments

compnerd (approved):

PR #662, opened 2023-10-04
Azure: publish cab files alongside with msi's

lxbndr:
Had to move msi and cab files to dedicated dirs before publishing, as Azure only can publish whole directories or single files.


lxbndr:
Had success on my local agent: https://dev.azure.com/lxbndr/swift-windows/_build/results?buildId=379&view=results

compnerd (approved):

PR #663, opened 2023-10-10
build.ps1: wait for msiexec completion

lxbndr:
msiexec actually runs the setup process in the background, and there is a possibility of silent failure when unpacking SDK and Toolchain into the same location simultaneously.

`Start-Process` could wait for each step to finish, which is what we want here. I didn't really want to introduce `Start-Process` into the `Invoke-Program`, as it makes things a bit more complicated and unpredictable (mostly because `Start-Process` requires additional quoting for arguments with spaces). Also, we're not handling errors here, but `Invoke-Program` doesn't catch msiexec background errors either.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -269,6 +269,7 @@ function Invoke-Program() {
     [Parameter(Position = 0, Mandatory = $true)]
     [string] $Executable,
     [switch] $OutNull = $false,
+    [switch] $Wait = $false,
     [string] $OutFile = "",
     [Parameter(Position = 1, ValueFromRemainingArguments)]
     [string[]] $Args
@@ -276,7 +277,11 @@ function Invoke-Program() {
 
   if ($ToBatch) {
     # Print the invocation in batch file-compatible format
-    $OutputLine = "`"$Executable`""
+    if ($Wait) {
+      $OutputLine = "start `"`" /wait `"$Executable`""
+    } else {
+      $OutputLine = "`"$Executable`""
+    }
     $ShouldBreakLine = $false
     for ($i = 0; $i -lt $Args.Length; $i++) {
       if ($ShouldBreakLine -or $OutputLine.Length -ge 40) {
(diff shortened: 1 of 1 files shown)


compnerd (commented):
@tristanlabelle is there a way to support the `-Wait` with the `Invoke-Program`?  @lxbndr the Invoke-Program is important to supporting `-ToBatch`

lxbndr:
@compnerd I see... It seems possible, we just have to recheck everything and make sure everything works. The first thing I noticed when replaced `& ...` invocation in `Invoke-Program` with `Start-Process`, is that plist generation is broken. It has that large string of python code as argument, and `Start-Process` breaks it into separate arguments. That behavior is expected and documented. We can add more code to process arguments and add quotes where needed. Also, error processing needs to be altered.
Or maybe @tristanlabelle could suggest a decent solution👍

lxbndr:
Oh, and `-NoNewWindow` flag should be used, I guess. We don't need multiple windows popping out.

lxbndr:
Phew, almost forgot about error handling. Not it should be good.

I introduced `-Wait` parameter for `Invoke-Program`. Inside it is implemented as a separate branch, because I wanted to preserve classic calls for most of invocations. `-ToBatch` is fully supported.

While `-Wait` is technically has no difference for classic blocking programs (and it really works, I tested it with a row of CMake invocations), it seems to be overkill to invoke everything with `Start-Process`. Also, it has some caveats:
- Looks like it is a bit slower than a direct call. It adds ~1s of extra time after the program finishes on my tests. Every time.
- It waits for all child processes to finish, which is not always the desirable behavior. I observed a hang once or twice, when "Microsoft Program Database" child process was left alive after CMake configuration phase. I guess it is detached, but still, `Start-Process` was blocked until I killed that process manually.

But for cases when we are know what we're waiting for it is a right tool to use.

build.ps1
```diff
@@ -257,14 +257,19 @@ function Invoke-Program() {
...
     [string] $OutFile = "",
     [Parameter(Position = 1, ValueFromRemainingArguments)]
     [string[]] $Args
   )
 
   if ($ToBatch) {
     # Print the invocation in batch file-compatible format
-    $OutputLine = "`"$Executable`""
+    if ($Wait) {
+      $OutputLine = "start `"$(Split-Path $Executable -leaf)`" /wait `"$Executable`""
```
compnerd, line 269: Why the leaf path only?  It seems that we should be able to use the full path right?  I'm concerned about the command not being in the path.

compnerd (approved):
Seems reasonable to me.  @tristanlabelle any comments?

build.ps1
lxbndr: There is such thing as "title" in the [parameters](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/start) of `start` command. And `start` treats first quoted parameter as a title. As we're quoting executable already, it becomes "a title", and then other arguments are treated as `start` options... A 
complete mess. I tried some alternative forms (like quoting whole command string along with options), and that didn't help. Seems like the title could be omitted only if there is no quotes at all.

Basically it is just an arbitrary string, and it could be empty quotes. I just thought it makes sense to provide some readable value, but not too long (i.e. not the full path). Maybe to avoid confusion and excessive code, I just have to use "" here. WDYT about ""?

compnerd: That might work? this seems very confusing 😓

lxbndr: Yeah, that actually works.
```
:: ✅ Works, no quotes
> start /wait C:\Windows\System32\msiexec.exe /qn /a a1 TARGETDIR=C:\Downloads\T

:: ✅ Works, the executable is parsed before the quoted argument
> start /wait C:\Windows\System32\msiexec.exe /qn /a "a1" TARGETDIR=C:\Downloads\T

:: 🚫 Not works, quoted part is parsed as title, and then Start looks for a valid option or executable
> start /wait "C:\Windows\System32\msiexec.exe" /qn /a a1 TARGETDIR=C:\Downloads\T
Invalid switch - "/qn".

:: 🚫 Not works. Opens a new cmd instance with the "title" from quoted part
> start /wait "C:\Windows\System32\msiexec.exe /qn /a a1 TARGETDIR=C:\Downloads\T"

:: ✅ Works, as the "title" is found empty
> start /wait "" "C:\Windows\System32\msiexec.exe" /qn /a "a1" TARGETDIR="C:\Downloads\T"
```

This weird parsing behavior is seems to be well known, and is mentioned on [ss64.com](https://ss64.com/nt/start.html) (my favorite source of advanced info and best practices about cmd in particular):
>Always include a TITLE this can be a simple string like "My Script" or just a pair of empty quotes ""
According to the Microsoft documentation, the title is optional, but depending on the other options chosen you can have problems if it is omitted.

lxbndr:
@compnerd I replaced confusing code with empty quotes, as it looks less noisy to me.

build.ps1
```diff
@@ -304,12 +309,32 @@ function Invoke-Program() {
...
+      } elseif ("" -ne $OutFile) {
+        $Process = Start-Process -Wait -PassThru -NoNewWindow -RedirectStandardOutput $OutFile -FilePath $Executable -ArgumentList $QuotedArgs
+      } else {
+        $Process = Start-Process -Wait -PassThru -NoNewWindow -FilePath $Executable -ArgumentList $QuotedArgs
+      }
+      $LastExitCode = $Process.ExitCode
     } else {
-      & $Executable @Args
+      if ($OutNull) {
+        $ $Executable @Args | Out-Null
```
compnerd, line 332: Hmm, what does `$` do?

lxbndr:
Looks like that ship has sailed

PR #667, opened 2023-10-19
GHA: package the VC++ Runtime in the installer

compnerd:
As we prepare to enable the ARM64 toolchain installer, we need to package the runtime as winget does not provide a means to do target dependent requirements.


compnerd:
tested at: https://github.com/thebrowsercompany/swift-build/actions/runs/6566663451

PR #668, opened 2023-10-19
AzDO: use pinned toolchain for early swift-syntax build

lxbndr:
Also:
- bump dependencies (curl 8.4.0, swift-asn1, swift-certificates, swift-collections, swift-crypto)
- specify `SWIFT_CLANG_LOCATION` unconditionally (not only for arm64 builds)
- adjust VS redist packaging

### DIFF
--- a/.azure/vs2022.yml
+++ b/.azure/vs2022.yml
@@ -40,17 +40,17 @@ resources:
     - repository: apple/swift-asn1
       endpoint: GitHub
       name: apple/swift-asn1
-      ref: refs/tags/0.8.0
+      ref: refs/tags/1.0.0
       type: github
     - repository: apple/swift-certificates
       endpoint: GitHub
       name: apple/swift-certificates
-      ref: refs/tags/0.6.0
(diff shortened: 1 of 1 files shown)


lxbndr:
Running a final build check right now, but as it is so long on a cloud agent, I see no reason to postpone the review. An early feedback would not harm.

compnerd (approved):

lxbndr:
Oh well... When everything seems to be alright, Python decides he doesn't get enough attention. Azure agent has Python 3.12 installed, and FindPython3 prefers its interpreter even when 3.9 is in the path. "Building LLDB Python wrapper" fails with script error "ModuleNotFoundError: No module named 'distutils'".

I will fix that and update PR after checks.

lxbndr:
Ok, I think this is safe to land. Even though I still can't get past pipeline limits (max I got it ~95% of the build), Python config in the CMake logs seems to be ok.
@compnerd

compnerd (approved):

PR #677, opened 2023-10-31
build: ensure that the temporary directory exists

compnerd:
Create the temporary directory before extracting the burn engine which would otherwise fail.


tristanlabelle (approved):
Nothing objectionable here

PR #681, opened 2023-11-02
build: remove build script

compnerd:
This has been migrated into the apple/swift repository. There is no reason to have a fork of the script.


tristanlabelle (approved):
Amen!

PR #682, opened 2023-11-09
GHA: adjust the compilers build for macros

compnerd:
In order to enable macros, we need to have the previous toolchain installed.  This should be able to provide the necessary compiler for ARM64 cross-compilations as well, and allow us to reduce the setup requirements. Take the opportunity to re-use the prebuilt clang universally.

### DIFF
--- a/.github/workflows/swift-toolchain.yml
+++ b/.github/workflows/swift-toolchain.yml
@@ -476,7 +476,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        arch: ['amd64', 'arm64']
+        include:
+          - arch: 'amd64'
+            cpu: 'x86_64'
+            triple: 'x86_64-unknown-windows-msvc'
+
+          - arch: 'arm64'
+            cpu: 'aarch64'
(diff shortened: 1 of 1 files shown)


.github/workflows/swift-toolchain.yml
```diff
@@ -1020,16 +1042,21 @@ jobs:
                 -D SWIFT_PATH_TO_SWIFT_SYNTAX_SOURCE=${{ github.workspace }}/SourceCache/swift-syntax `
                 -D SWIFT_PATH_TO_STRING_PROCESSING_SOURCE=${{ github.workspace }}/SourceCache/swift-experimental-string-processing
       - name: Build Swift Standard Library
-        run: cmake --build ${{ github.workspace }}/BinaryCache/swift
+        run: |
+          Remove-Item env:\SDKROOT
```
tristanlabelle, line 1046: I wonder if we can write `SDKROOT=` to `GITHUB_ENV` to remove the variable once and for all?
(in a future change)

tristanlabelle (approved):

PR #693, opened 2023-12-15
Update the 5.10 release branch

hyp:


compnerd:
I don't understand the revert.  That effectively disables macros doesn't it?

hyp:
> I don't understand the revert. That effectively disables macros doesn't it?

This caused the following failure, so I disabled this specific change:

```
Run $module = "C:\a\swift-build\swift-build/BinaryCache/1/cmake/modules/SwiftSyntaxConfig.cmake"
Get-Content: C:\a\_temp\4742a4db-1ae7-4860-91b8-d9d7235cc3c4.ps1:4
Line |
   4 |  (Get-Content $module).Replace("${bindir}", '<BINARY_DIR>') | Set-Cont …
     |   ~~~~~~~~~~~~~~~~~~~
     | Cannot find path 'C:\a\swift-build\swift-build\BinaryCache\1\cmake\modules\SwiftSyntaxConfig.cmake' because it
     | does not exist.
```

Should I bring back that commit?

compnerd:
Hmm, I believe that change should be in the 5.10 release, so this seems like something to investigate (or we might have regressed on 5.10 with the macros support 🤦‍♂️).

hyp:
> Hmm, I believe that change should be in the 5.10 release, so this seems like something to investigate (or we might have regressed on 5.10 with the macros support 🤦‍♂️).

cool, let me see if I can dig into it more.

hyp:
Okay, good news, my swift 5.10 cherry-pick already solves this, so I can revert the revert :D

hyp:
@compnerd it's ready now, but I need to merge https://github.com/apple/swift/pull/70500 first

.github/workflows/swift-toolchain.yml
```diff
@@ -73,7 +73,14 @@ jobs:
           sudo DEBIAN_FRONTEND=noninteractive apt-get -qq update -yq
           sudo DEBIAN_FRONTEND=noninteractive apt-get -qq -o Dpkg::Use-Pty=0 install -yq repo libxml2-utils
 
-          repo init --quiet --groups default --depth 1 -u https://github.com/compnerd/swift-build -b main
+          # Which branch is this workflow based on
+          branch_version=main
+          if [[ "$branch_version" == "main" ]] ; then
+            branch_name="main"
+          else
+            branch_name="release/${branch_version}"
```
compnerd, line 81: I think that we should be able to use the github variable for the branch names.  We follow the same branch names apple does.  This means that we should just drop the `release/` prefix on the branch name.

compnerd (approved):

hyp:
@compnerd looks like this branch is protected and even a rebased merge isn't working here

PR #695, opened 2023-12-19
GHA: update the {download,upload}-artifact action

compnerd:
Use the new V4 release of the artifact which should have performance improvements though it will change some of the behaviour.


compnerd:
Tested at https://github.com/thebrowsercompany/swift-build/actions/runs/7253943547

hjyamauchi (approved):

PR #700, opened 2024-01-04
GHA: bump sccache to 0.7.4

compnerd:
Use the forked ccache-action to use the latest sccache release. This should allow us to use the direct mode and hopefully get slightly faster builds.


hyp (approved):
Thanks!

PR #702, opened 2024-01-08
Add a workflow to build Swift 5.10 release on schedule

hyp:
The new workflow reuses the `swift-toolchain.yml` workflow. I tested this workflow at thebrowsercompany/swift-build .

The release workflow builds releases, like Swift 5.10 nightly. As a follow-up, I will add additional logic to not rebuild the release when the toolchain checkout doesn't change.

### DIFF
--- a/.github/workflows/release-swift-toolchain-schedule.yml
+++ b/.github/workflows/release-swift-toolchain-schedule.yml
@@ -1,0 +1,16 @@
+name: Build release Swift toolchains
+
+on:
+  workflow_dispatch:
+  # Schedule to build a new release toolchain nightly.
+  schedule:
+    - cron: "10 0 * * */1"
+
+jobs:
+  # Each job builds a release toolchain for a specific Swift version.
+  build-release-5_10:
+    uses: compnerd/swift-build/.github/workflows/swift-toolchain.yml@release/5.10
+    secrets:
+      SYMBOL_SERVER_PAT: ${{ secrets.SYMBOL_SERVER_PAT }}
+      CERTIFICATE: ${{ secrets.CERTIFICATE }}
+      PASSPHRASE: ${{ secrets.PASSPHRASE }}
--- a/.github/workflows/release-swift-toolchain-schedule.yml
+++ b/.github/workflows/release-swift-toolchain-schedule.yml
@@ -9,6 +9,8 @@ on:
 jobs:
   # Each job builds a release toolchain for a specific Swift version.
   build-release-5_10:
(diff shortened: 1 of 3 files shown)


.github/workflows/release-swift-toolchain-schedule.yml
```diff
@@ -0,0 +1,16 @@
...
+on:
+  workflow_dispatch:
+  # Schedule to build a new release toolchain nightly.
+  schedule:
+    - cron: "10 0 * * */1"
+
+jobs:
+  # Each job builds a release toolchain for a specific Swift version.
+  build-release-5_10:
+    uses: compnerd/swift-build/.github/workflows/swift-toolchain.yml@release/5.10
```
compnerd, line 12: Why does it need the `compnerd/swift-build` prefix? Should `.github/workflows/swift-toolchain.yml@release/5.10` not be repository relative?

.github/workflows/swift-toolchain.yml
```diff
@@ -25,6 +25,33 @@ on:
         description: 'Code Sign'
         default: false
         type: boolean
+  workflow_call:
+    inputs:
+      snapshot:
+        description: 'Build Swift at a tagged snapshot'
+        default: false
+        type: boolean
```
compnerd, line 33: Hmm, we really should see if it is possible to build at a tagged snapshot at this point. I think that with the context step, it no longer is, and it might make sense to just drop this option. But that can be a separate change.

.github/workflows/release-swift-toolchain-schedule.yml
hyp: Good question, let me check

hyp: @compnerd unfortunately GitHub has a limitation for local workflow reference - I can't specify another branch:

```
invalid value workflow reference: cannot specify version when calling local workflows
```

So it looks like I have to use the owner/repo prefix :/

compnerd: I wonder if we can use something from the github context to get the owner/repo (i.e. `${{ github.repository }}` instead of `compnerd/swift-build`.  If not, lets add a comment explaining why this is hardcoded.

hyp: Unfortunately that doesn't work:
```
invalid value workflow reference: references to workflows must be prefixed with format 'owner/repository/' or './' for local workflows
```

I added a note now in the updated patch.

compnerd (approved):
That cleanup on the swift-toolchain.yml though ... 👌

PR #706, opened 2024-01-18
Add sourcekit-lsp initialization smoke test

z2oh:
Not sure of the best place to add the initialization request message. Because of the specific format of the message (must be CLRF; must be precise byte-length) it seemed difficult to inline into the GHA. I added a new `tests` dir with the file there, but I am reluctant to clutter the repo root. Is there a better place to put this?

### DIFF
--- a/.github/workflows/swift-toolchain.yml
+++ b/.github/workflows/swift-toolchain.yml
@@ -2306,6 +2306,16 @@ jobs:
       - run: swift test -Xswiftc -DENABLE_TESTING
         working-directory: ${{ github.workspace }}/SourceCache/swift-win32
 
+      - name: Test that sourcekit-lsp initializes
+        working-directory: ${{ github.workspace }}/SourceCache/swift-win32
+        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/swift-toolchain.yml
```diff
@@ -2306,6 +2306,16 @@ jobs:
       - run: swift test -Xswiftc -DENABLE_TESTING
         working-directory: ${{ github.workspace }}/SourceCache/swift-win32
 
+      - name: Test that sourcekit-lsp initializes
```
compnerd, line 2309: I would just do `SourceKit Validation`

```suggestion
      - name: SourceKit-LSP Validation
```

tests/sourcekit-lsp-initialization-message
```diff
@@ -0,0 +1,11 @@
...
+
+{
+    "jsonrpc": "2.0",
+    "id": 0,
+    "method": "initialize",
+    "params": {
+        "capabilities": {},
+        "trace": "verbose",
+    }
+}
```
compnerd, line 11: Missing newline at the end of the file?

z2oh: I'll have to test if including one breaks the test. There's not supposed to be a newline at the end of the RPC message

PR #717, opened 2024-02-19
Work around the ARM64 libdispatch miscompile.

hjyamauchi:


compnerd (commented):
This feels pretty drastic. Can we try to isolate this down a little bit?

hjyamauchi:
Yep, plan to.

compnerd:
@hjyamauchi is this still needed?

PR #724, opened 2024-03-12
[Azure sym server] Upload the DLLs

sebmarchand:
Upload the PE files generated by the ArcCore build to the Azure symbol server in addition to the PDBs.

Also make the symbol search pattern explicit as the default wasn't gathering all desired files.

### DIFF
--- a/.github/workflows/swift-toolchain.yml
+++ b/.github/workflows/swift-toolchain.yml
@@ -1,8 +1,6 @@
 name: Development Snapshot
 
 on:
-  schedule:
-    - cron: "0 */6 * * *"
   workflow_dispatch:
     inputs:
       snapshot:
@@ -21,7 +19,7 @@ on:
 
       debug_info:
         description: 'Emit PDBs (Debug Info)'
-        default: true
+        default: false
         type: boolean
       signed:
         description: 'Code Sign'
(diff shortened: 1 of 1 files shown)


.github/workflows/swift-toolchain.yml
```diff
@@ -793,12 +793,22 @@ jobs:
           path: |
             ${{ github.workspace }}/BinaryCache/1/**/*.pdb
 
-      - uses: microsoft/action-publish-symbols@v2.1.6
+      - name: Upload PDBs to Azure
```
tristanlabelle, line 796: This is in the compilers job. I believe we need this in the sdk job (or both)

compnerd: Yes, we will want it in both. I'd really like to try to get the glob working though.

.github/workflows/swift-toolchain.yml
```diff
@@ -793,12 +793,32 @@ jobs:
...
+        uses: microsoft/action-publish-symbols@v2.1.6
         if: ${{ needs.context.outputs.debug_info }}
         with:
           accountName: ${{ vars.SYMBOL_SERVER_ACCOUNT }}
           personalAccessToken: ${{ secrets.SYMBOL_SERVER_PAT }}
           symbolsFolder: ${{ github.workspace }}/BinaryCache/1
+          searchPattern: '**/bin/**/*.pdb'
```
tristanlabelle, line 803: @compnerd I made these explicit (from the action's default). Is this pattern correct, and should I follow the same format for dlls and exes? I don't have a toolchain build handy to check.

.github/workflows/swift-toolchain.yml
tristanlabelle: Pushed a fix.

.github/workflows/swift-toolchain.yml
tristanlabelle: Turns out `dispatch.dll`/`dispatch.pdb` are not emitted within a "bin" directory, so we are not uploading them. (fyi @hjyamauchi )

tristanlabelle:
Trial run here: https://github.com/thebrowsercompany/swift-build/actions/runs/8332881460

compnerd (commented):
I'm fine with this conceptually, but I think that I want to see if we can collapse these into a single run. That would reduce the overall impact of the step as we can upload more items in parallel/

.github/workflows/swift-toolchain.yml
compnerd: Yeah, libdispatch not using `bin` for the output is why the default pattern didn't work.

compnerd:
Doing a collapsed variant run at https://github.com/thebrowsercompany/swift-build/actions/runs/8333794115

compnerd (approved):
Well, this is unfortunate, both forms fail. We can take a look at this later I suppose, merging this to avoid blocking.

tristanlabelle:
@compnerd We can't use a single invocation because https://github.com/microsoft/action-publish-symbols/issues/32

PR #729, opened 2024-03-29
Clarify install instructions in WindowsQuickStart.md

kendalharland:


docs/WindowsQuickStart.md
```diff
@@ -31,11 +37,14 @@ del /q vs_community.exe
...
 
+> [!INFO] This step only needs to be completed if your User is not an Administrator, as
+> Adminstrators already have permission to create symbolic links.
+
 Grant your user the `SeCreateSymbolicLinkPrivilege` rights.  This can be done by
 applying a Group Policy Object to the system.  Run `gpedit.msc` and navigate to
 
 ~~~
-Computer Configuration\Windows Settings\Security Settings\Local Policies\User Rights Assignment
+Computer > Configuration\Windows > Settings\Security > Settings\Local > Policies\User Rights Assignment
```
compnerd, line 47: Hmm, the path itself is the path which shows up in the UI. I don't have an issue with the `>` usage, but then we should change the `\` to the appropriate `/`. Is the intermediate setting not "Local Policies", rather than "Policies" rooted under "Local"?

hjyamauchi: Yeah how about putting `>` where the `\`'s were?

kendalharland: Whoops, that's what I meant to do :P thanks for correcting. Done.

compnerd (approved):

PR #730, opened 2024-03-29
Kendal/devin 1170 build swift inspect in the swift build GitHub workflows

kendalharland:


default.xml
```diff
@@ -29,6 +29,7 @@
   <project remote="github" name="apple/swift-format" path="swift-format" />
   <project remote="github" name="apple/swift-foundation" path="swift-foundation" />
   <project remote="github" name="apple/swift-installer-scripts" path="swift-installer-scripts" />
+  <project remote="github" name="apple/swift-inspect" path="swift-inspect" />
```
compnerd, line 32: This seems wrong and will break the checkout - there is no such repository. swift-inspect is a tool which resides in apple/swift.

.github/workflows/swift-toolchain.yml
```diff
@@ -1718,6 +1726,35 @@ jobs:
...
+                  -D CMAKE_Swift_FLAGS_RELEASE="-O" `
+                  -D CMAKE_SYSTEM_NAME=Windows `
+                  -D CMAKE_SYSTEM_PROCESSOR=${{ matrix.cpu }} `
+                  -G Ninja `
+                  -S ${{ github.workspace }}/SourceCache/swift-llbuild `
+                  -D LLBUILD_SUPPORT_BINDINGS=Swift `
+                  -D SQLite3_LIBRARY=${{ github.workspace }}/BuildRoot/Library/sqlite-3.43.2/usr/lib/SQLite3.lib `
+                  -D SQLite3_INCLUDE_DIR=${{ github.workspace }}/BuildRoot/Library/sqlite-3.43.2/usr/include
+      - name: Build swift-inspect
+        run: cmake --build ${{ github.workspace }}/BinaryCache/swift-inspect
```
compnerd, line 1756: The swift compiler is not specified, and given that we want to build with the just built swift compiler, this is going to require some additional tweaking.

compnerd (changes requested):

compnerd (changes requested):
This doesn't work. We cannot build swift inspect before we build the SDK as it is sworn in swift and requires the just built swift compiler and SDK.

.github/workflows/swift-toolchain.yml
```diff
@@ -2118,9 +2118,118 @@ jobs:
           symbolsFolder: ${{ github.workspace }}/BinaryCache
           searchPattern: '**/*.exe'
 
+  debugging_tools:
+    name: Debugging Tools
+    environment: "kendal/testing"
```
compnerd, line 2123: Debugging left overs?

.github/workflows/swift-toolchain.yml
```diff
@@ -2118,9 +2118,118 @@ jobs:
...
+
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - arch: amd64
+            cpu: x86_64
+            triple: x86_64-unknown-windows-msvc
+
+          # TODO(125): Fix the arm64 build and uncomment this.
```
compnerd, line 2137: Please qualify the issue

```suggestion
          # TODO(thebrowsercompany/swift-build#125): Fix the arm64 build and uncomment this.
```

.github/workflows/swift-toolchain.yml
```diff
@@ -2118,9 +2118,118 @@ jobs:
...
+      - name: Download devtools
+        uses: actions/download-artifact@v4
+        with:
+          name: devtools-${{ matrix.arch }}
+          path: ${{ github.workspace }}/BuildRoot/Library
+
+      - name: Download cmark-gfm-${{ matrix.arch }}
+        uses: actions/download-artifact@v4
+        with:
+          name: cmark-gfm-${{ matrix.arch }}-0.29.0.gfm.13
```
compnerd, line 2158: This is for the runtime, this should always be `amd64`.

default.xml
kendalharland: Updated

.github/workflows/swift-toolchain.yml
```diff
@@ -2118,9 +2118,118 @@ jobs:
...
+        run: |
+          echo ${env:TOOLCHAIN_PATH} | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
+
+      - uses: compnerd/gha-setup-vsdevenv@main
+        with:
+          host_arch: amd64
+          components: 'Microsoft.VisualStudio.Component.VC.Tools.x86.x64;Microsoft.VisualStudio.Component.VC.Tools.ARM64'
+          arch: ${{ matrix.arch }}
+
+      # TODO(50): Workaround a file layout bug in the Windows SDK.
```
compnerd, line 2199:
```suggestion
      # TODO(thebrowsercompany/swift-build#50): Workaround a file layout bug in the Windows SDK.
```

.github/workflows/swift-toolchain.yml
```diff
@@ -2118,9 +2118,118 @@ jobs:
...
+          # TODO(kendal): Figure out why Powershell won't execute this command when it's split into multiple lines.
+          swift build --scratch-path="${{ github.workspace }}/BinaryCache/${{ matrix.arch }}/swift-inspect" --sdk="${env:SDKROOT}" --package-path="${{ github.workspace }}/SourceCache/swift/tools/swift-inspect" -c="release" -Xbuild-tools-swiftc="-I${env:SDKROOT}/usr/lib/swift" -Xbuild-tools-swiftc="-L${env:SDKROOT}/usr/lib/swift/windows" -Xcc="-I${env:SDKROOT}/usr/lib/swift" -Xlinker="-L${env:SDKROOT}/usr/lib/swift/windows" -debug-info-format="none" -Xcc="-I${env:SDKROOT}/usr/include/swift/SwiftRemoteMirror" -Xlinker="${env:SDKROOT}/usr/lib/swift/windows/${{ matrix.cpu }}/swiftRemoteMirror.lib"
+
+      - name: Copy swift-inspect executable
+        id: copy-swift-inspect-exe
+        run: |
+          New-Item -Path ${env:TOOLCHAIN_PATH} -ItemType Directory -Force | Out-Null
+          Copy-Item ${{ github.workspace }}/BinaryCache/${{ matrix.arch }}/swift-inspect/x86_64-unknown-windows-msvc/release/swift-inspect.exe -Destination ${env:TOOLCHAIN_PATH}
```
compnerd, line 2223: This seems wrong ... we want to copy the binary into the artifact path which we upload below. We don't want to upload the full staged toolchain image.

.github/workflows/swift-toolchain.yml
```diff
@@ -2129,16 +2238,28 @@ jobs:
         arch: ['amd64' , 'arm64']
 
     steps:
+      - name: Download Debugging Tools
+        uses: actions/download-artifact@v4
+        # TODO(125): Fix the arm64 build and uncomment this.
```
compnerd, line 2243:
```suggestion
        # TODO(thebrowsercompany/swift-build#125): Fix the arm64 build and uncomment this.
```

.github/workflows/swift-toolchain.yml
```diff
@@ -2118,9 +2118,119 @@ jobs:
...
           searchPattern: '**/*.exe'
 
+  debugging_tools:
+    name: Debugging Tools
+    needs: [context, compilers, devtools, sdk]
+    runs-on: ${{ needs.context.outputs.windows_build_runner }}
+    env:
+      TOOLCHAIN_PATH: ${{ github.workspace }}/BuildRoot/Library/Developer/Toolchains/unknown-Asserts-development.xctoolchain/usr/bin
+      # Where to store new binaries.
+      DEBUGGING_TOOLS_DIR: ${{ github.workspace }}/BuildRoot-DebuggingTools/Library/Developer/Toolchains/unknown-Asserts-development.xctoolchain/usr
```
compnerd, line 2128: This is different from the other portions of the workflow. Should we try to keep this uniform and then do a change universally? How would this scale when you have $n$ repositories to build in a stage? (e.g. devtools).

.github/workflows/swift-toolchain.yml
```diff
@@ -2118,9 +2118,119 @@ jobs:
...
+        uses: actions/download-artifact@v4
+        with:
+          name: cmark-gfm-amd64-0.29.0.gfm.13
+          path: ${{ github.workspace }}/BuildRoot/Library/cmark-gfm-0.29.0.gfm.13/usr
+
+      - name: Download ${{ matrix.arch }} Windows Runtime
+        uses: actions/download-artifact@v4
+        with:
+          name: windows-sdk-${{ matrix.arch }}
+          path: ${{ github.workspace }}/BinaryCache
```
compnerd, line 2166: Do we not need the runtime for AMD64 always? We have a dependency on it for the toolchain I thought.

.github/workflows/swift-toolchain.yml
```diff
@@ -2118,9 +2118,119 @@ jobs:
...
+          mv ${env:SDKROOT}/usr/lib/swift/windows/FoundationXML.lib ${env:SDKROOT}/usr/lib/swift/windows/${{ matrix.cpu }}/
+          mv ${env:SDKROOT}/usr/lib/swift/windows/FoundationNetworking.lib ${env:SDKROOT}/usr/lib/swift/windows/${{ matrix.cpu }}/
+
+      - name: Build swift-inspect
+        run: |
+          $env:SDKROOT="${{ github.workspace }}/BuildRoot/Library/Developer/Platforms/Windows.platform/Developer/SDKs/Windows.sdk"
+          # TODO(kendal): Figure out why Powershell won't execute this command when it's split into multiple lines.
+          swift build --scratch-path="${{ github.workspace }}/BinaryCache/${{ matrix.arch }}/swift-inspect" --sdk="${env:SDKROOT}" --package-path="${{ github.workspace }}/SourceCache/swift/tools/swift-inspect" -c="release" -Xbuild-tools-swiftc="-I${env:SDKROOT}/usr/lib/swift" -Xbuild-tools-swiftc="-L${env:SDKROOT}/usr/lib/swift/windows" -Xcc="-I${env:SDKROOT}/usr/lib/swift" -Xlinker="-L${env:SDKROOT}/usr/lib/swift/windows" -debug-info-format="none" -Xcc="-I${env:SDKROOT}/usr/include/swift/SwiftRemoteMirror" -Xlinker="${env:SDKROOT}/usr/lib/swift/windows/${{ matrix.cpu }}/swiftRemoteMirror.lib"
```
compnerd, line 2218: Are the `=` after `-Xcc` and `-Xlinker` required?

.github/workflows/swift-toolchain.yml
```diff
@@ -2129,16 +2239,28 @@ jobs:
...
+          path: ${{ github.workspace }}/BuildRoot/Library
+
       - name: Download Compilers
         uses: actions/download-artifact@v4
         with:
           name: compilers-${{ matrix.arch }}
           path: ${{ github.workspace }}/BuildRoot/Library
-      - uses: actions/download-artifact@v4
+
+      - name: Download Devtools
```
compnerd, line 2256: I think that the stage might be `DevTools`.

.github/workflows/swift-toolchain.yml
kendalharland: Hmm, I'm not sure what you're asking: are you saying that we should use `BuildRoot-DevTools` here instead of `BuildRoot-DebuggingTools`, or that this particular job should not store all of the tools it builds in a single common root directory, specified by this env var?

.github/workflows/swift-toolchain.yml
kendalharland: Oh yes, thanks for catching. This is just a bad copy-paste. I got lucky here that this is currently always evaluating `${{ matrix.arch }}` to `amd64` and didn't notice.

.github/workflows/swift-toolchain.yml
kendalharland: Seems like it (See: https://github.com/thebrowsercompany/swift-build/actions/runs/8623456207/job/23636742664). I'll have to debug this while figuring out why this command doesn't work when split across multiple lines. The solution might just be to move the command to a separate powershell script in a later patch.

.github/workflows/swift-toolchain.yml
kendalharland: Fixed capitalization

.github/workflows/swift-toolchain.yml
compnerd: I'm simply pointing out the use of the environment variable is different.

.github/workflows/swift-toolchain.yml
compnerd: I'd rather not externalise this into a script.

compnerd:
I wonder if this also fixes the ARM64 build (we would be missing libraries previously)

.github/workflows/swift-toolchain.yml
kendalharland: Ah, I see. Doing it later, all at once sounds fine with me.

kendalharland:
> I wonder if this also fixes the ARM64 build (we would be missing libraries previously)

It's possible! I'll make another happy-path PR on top if see it resolves the arm64 failure

.github/workflows/swift-toolchain.yml
kendalharland: SGTM. Will update after I debug.

kendalharland:
> This must happen at the end of the build but before the packaging steps. This is the wrong place as the swift compiler has not yet been built.

I'm not sure how to make Github resolve this comment :_( it still shows up under "Changes requested". any idea?

.github/workflows/swift-toolchain.yml
```diff
@@ -2118,9 +2118,117 @@ jobs:
...
+          # TODO(thebrowsercompany/swift-build#125): Fix the arm64 build and uncomment this.
+          # - arch: arm64
+          #   cpu: aarch64
+          #   triple: aarch64-unknown-windows-msvc
+
+    steps:
+      - name: Download compilers
+        uses: actions/download-artifact@v4
+        with:
+          name: compilers-${{ matrix.arch }}
```
compnerd, line 2145: Do we not need the amd64 compilers always as we are cross-compiling x64 -> arm64 and x64 -> x64?

.github/workflows/swift-toolchain.yml
```diff
@@ -2118,9 +2118,117 @@ jobs:
...
+          path: ${{ github.workspace }}/BuildRoot/Library
+
+      - name: Download devtools
+        uses: actions/download-artifact@v4
+        with:
+          name: devtools-${{ matrix.arch }}
```
compnerd, line 2151: Do we not need the amd64 devtools always as we are cross-compiling x64 -> arm64 and x64 -> x64?

.github/workflows/swift-toolchain.yml
```diff
@@ -2205,7 +2325,8 @@ jobs:
               -p:PASSPHRASE=${{ secrets.PASSPHRASE }} `
               -p:DEVTOOLS_ROOT=${{ github.workspace }}/BuildRoot/Library/Developer/Toolchains/unknown-Asserts-development.xctoolchain `
               -p:TOOLCHAIN_ROOT=${{ github.workspace }}/BuildRoot/Library/Developer/Toolchains/unknown-Asserts-development.xctoolchain `
-              -p:SWIFT_INSPECT_BUILD=false `
+              -p:INCLUDE_SWIFT_INSPECT=${{ matrix.arch == 'amd64' && true || false }} `
+              -p:SWIFT_INSPECT_BUILD=${{ github.workspace }}/BuildRoot/Library/Developer/Toolchains/unknown-Asserts-development.xctoolchain/usr `
```
compnerd, line 2329: Really? This is the path to the build?

compnerd (commented):
This looks pretty close!

.github/workflows/swift-toolchain.yml
kendalharland: I believe so. Mind if I iterate on this while enabling the arm64 build in the next PR? I need to figure out how to cross compile with SPM and it'll be easier to iterate with a smaller diff base.

.github/workflows/swift-toolchain.yml
kendalharland: Same here, mind if I address this while investigating issue # 125?

.github/workflows/swift-toolchain.yml
kendalharland: Updated to a directory outside of the directory we use for the staged tools.

.github/workflows/swift-toolchain.yml
compnerd: But this change would give you a smaller diff because we know that this should be unconditional.

.github/workflows/swift-toolchain.yml
compnerd: Likewise.

.github/workflows/swift-toolchain.yml
```diff
@@ -662,7 +662,7 @@ jobs:
           arch: ${{ matrix.arch }}
 
       - name: Install Swift Toolchain
-        uses: compnerd/gha-setup-swift@8f43ccc3e8bac89829862af09de9567c807c1c12
+        uses: compnerd/gha-setup-swift@main
```
kendalharland, line 665: Whoops, I got confused looking at the diffbase in our fork. I'll change this back

compnerd: The only difference should be the defaults, nothing else.

.github/workflows/swift-toolchain.yml
kendalharland: Er, I guess I could just download them both even if the subsequent steps aren't using them yet. I'll try this out tomorrow.

.github/workflows/swift-toolchain.yml
kendalharland: Done

.github/workflows/swift-toolchain.yml
kendalharland: Done

.github/workflows/swift-toolchain.yml
kendalharland: Reverted.

compnerd (approved):
While there is a small improvement to be made, I don't think its worth holding off on this change for that. The subsequent change to enable the ARM64 builds of swift-inspect would be an ideal opportunity to improve that.

PR #740, opened 2024-05-23
Add support for android SDK builds for the toolchain GHA workfow

hyp:


.github/workflows/swift-toolchain.yml
```diff
@@ -55,6 +55,7 @@ on:
 
 env:
   SCCACHE_DIRECT: yes
+  ANDROID_API_LEVEL: 28
```
compnerd, line 58: Do we have the space to move this into an input? That would make this configurable. I'm not a fan of using the environment to propagate settings.

.github/workflows/swift-toolchain.yml
```diff
@@ -179,6 +186,11 @@ jobs:
             echo CMAKE_Swift_FLAGS="-Xlinker -incremental:no -Xlinker -opt:ref -Xlinker -opt:icf" >> ${GITHUB_OUTPUT}
           fi
 
+          echo CMAKE_ANDROID_C_FLAGS="-ffunction-sections -fdata-sections -g" >> ${GITHUB_OUTPUT}
+          echo CMAKE_ANDROID_CXX_FLAGS="-ffunction-sections -fdata-sections -g" >> ${GITHUB_OUTPUT}
+          echo CMAKE_ANDROID_EXE_LINKER_FLAGS="" >> ${GITHUB_OUTPUT}
+          echo CMAKE_ANDROID_SHARED_LINKER_FLAGS="" >> ${GITHUB_OUTPUT}
```
compnerd, line 192: Should we consider sinking these into CMake caches or the swift build system?

.github/workflows/swift-toolchain.yml
```diff
@@ -200,6 +212,9 @@ jobs:
           echo windows_build_runner=${{ vars.WINDOWS_BUILD_RUNNER || 'windows-latest' }} >> ${GITHUB_OUTPUT}
           echo compilers_build_runner=${{ vars.COMPILERS_BUILD_RUNNER || vars.WINDOWS_BUILD_RUNNER || 'windows-latest' }} >> ${GITHUB_OUTPUT}
 
+          echo ANDROID_API_LEVEL=${{ env.ANDROID_API_LEVEL }} >> ${GITHUB_OUTPUT}
```
compnerd, line 215: This could be removed if we used the input right?

.github/workflows/swift-toolchain.yml
```diff
@@ -200,6 +212,9 @@ jobs:
...
+          echo WINDOWS_VISUAL_STUDIO_PATH="C:\Program Files\Microsoft Visual Studio\2022\Enterprise" >> ${GITHUB_OUTPUT}
```
compnerd, line 216: Please don't do this. We should query this using `vswhere`.

.github/workflows/swift-toolchain.yml
```diff
@@ -213,7 +228,64 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        arch: ['amd64', 'arm64', 'x86']
+        include:
+          - arch: amd64
```
compnerd, line 232: Names might be nice at this point - this is a pretty diverse set of configuration.

.github/workflows/swift-toolchain.yml
```diff
@@ -347,12 +423,77 @@ jobs:
...
+
+          - arch: arm64
+            BUILD_TOOLS: NO
+            BUILD_DATA: NO
+            cc: clang
+            cflags: -ffunction-sections -fdata-sections -g
+            cxx: clang++
+            cxxflags: -ffunction-sections -fdata-sections -g
+            os: Android
+            extra_flags: -DCMAKE_POSITION_INDEPENDENT_CODE=YES -D CMAKE_ANDROID_API=${{ needs.context.outputs.ANDROID_API_LEVEL }} -D CMAKE_ANDROID_ARCH_ABI=arm64-v8a "-DCMAKE_MAKE_PROGRAM=${{ needs.context.outputs.WINDOWS_VISUAL_STUDIO_PATH }}\Common7\IDE\CommonExtensions\Microsoft\CMake\Ninja\ninja.exe"
```
compnerd, line 464: I think that we should push down the default for `CMAKE_POSITION_INDEPENDENT_CODE` into the ICU CMakeLists.txt.

.github/workflows/swift-toolchain.yml
```diff
@@ -986,9 +1268,9 @@ jobs:
                 -D CURL_USE_LIBSSH2=NO `
                 -D CURL_USE_MBEDTLS=NO `
                 -D CURL_USE_OPENSSL=NO `
-                -D CURL_USE_SCHANNEL=YES `
+                -D CURL_USE_SCHANNEL=${{ matrix.windows_cmake_flag }} `
                 -D CURL_USE_WOLFSSL=NO `
-                -D CURL_WINDOWS_SSPI=YES `
+                -D CURL_WINDOWS_SSPI=${{ matrix.windows_cmake_flag }} `
```
compnerd, line 1273: We could use a ternary here instead - `matrix.os == 'Windows' && 'YES' || 'NO'`.

.github/workflows/swift-toolchain.yml
```diff
@@ -1006,8 +1288,8 @@ jobs:
                 -D USE_NGHTTP2=NO `
                 -D USE_NGTCP2=NO `
                 -D USE_QUICHE=NO `
-                -D USE_WIN32_IDN=YES `
-                -D USE_WIN32_LARGE_FILES=YES `
+                -D USE_WIN32_IDN=${{ matrix.windows_cmake_flag }} `
+                -D USE_WIN32_LARGE_FILES=${{ matrix.windows_cmake_flag }} `
```
compnerd, line 1292: Likewise

.github/workflows/swift-toolchain.yml
hjyamauchi: I have Community version of VS where this wouldn't work?

hjyamauchi:
It might be worth checking that this won't break the arm64 cross-compile build (`build.cmd -HostArchName=ARM64` on X64) or the Apple CI https://ci-external.swift.org/job/swift-main-windows-toolchain-arm64/ (though it's broken at the moment?)

.github/workflows/swift-toolchain.yml
tristanlabelle: Makes sense to me if `-ffunction-sections` and -fdata-sections` are required to build for Android, and I imagine we would keep `-g` here?

.github/workflows/swift-toolchain.yml
tristanlabelle, line 216: This is always used to derive the path to `ninja.exe`. We should do that only once and store it.

.github/workflows/swift-toolchain.yml
```diff
@@ -213,7 +228,64 @@ jobs:
...
+            cc: cl
+            cflags: ${{ needs.context.outputs.CMAKE_C_FLAGS }}
+            cxx: cl
+            cxxflags: ${{ needs.context.outputs.CMAKE_CXX_FLAGS }}
+            os: Windows
+            extra_flags: 
+
+          - arch: arm64
+            cc: clang
+            cflags: -ffunction-sections -fdata-sections -g
```
tristanlabelle, line 258: Derive from `CMAKE_ANDROID_C_FLAGS` / `CMAKE_ANDROID_CXX_FLAGS` to avoid repeating?

.github/workflows/swift-toolchain.yml
```diff
@@ -1192,46 +1623,74 @@ jobs:
...
+          $NDKHash = "A478D43D4A45D0D345CDA6BE50D79642B92FB175868D9DC0DFC86181D80F691E"
+          curl.exe -sL $NDKURL -o $env:TEMP\android-ndk-r26b-windows.zip
+          $SHA256 = Get-FileHash -Path $env:TEMP\android-ndk-r26b-windows.zip -Algorithm SHA256
+          if ($SHA256.Hash -ne $NDKHash) {
+            throw "NDK SHA256 mismatch ($($SHA256.Hash) vs $$NDKHash)"
+          }
+
+          Expand-Archive -Path $env:TEMP\android-ndk-r26b-windows.zip -DestinationPath ${{ github.workspace }}\BuildRoot\Library -Force
+          dir ${{ github.workspace }}\BuildRoot\Library
+          dir ${{ github.workspace }}\BuildRoot\Library\android-ndk-r26b
```
tristanlabelle, line 1639: Do we still need these dirs?

tristanlabelle (approved):
Thanks for iterating to get to this point, seems like it must have been challenging!

.github/workflows/swift-toolchain.yml
hyp: I set the name of the job using `name: ${{ matrix.os }} ${{ matrix.arch }} SQLite3`. Is that what you meant, or did you want to include a specific separate name in each matrix item?

.github/workflows/swift-toolchain.yml
hyp: Updated to derive where we need it using `vswhere`.

.github/workflows/swift-toolchain.yml
```diff
@@ -347,12 +423,77 @@ jobs:
...
+
+          - arch: armv7
+            BUILD_TOOLS: NO
+            BUILD_DATA: NO
+            cc: clang
+            cflags: -ffunction-sections -fdata-sections -g
+            cxx: clang++
+            cxxflags: -ffunction-sections -fdata-sections -g
+            os: Android
+            extra_flags: -DCMAKE_POSITION_INDEPENDENT_CODE=YES -D CMAKE_ANDROID_API=${{ needs.context.outputs.ANDROID_API_LEVEL }} -D CMAKE_ANDROID_ARCH_ABI=armeabi-v7a "-DCMAKE_MAKE_PROGRAM=${{ needs.context.outputs.WINDOWS_VISUAL_STUDIO_PATH }}\Common7\IDE\CommonExtensions\Microsoft\CMake\Ninja\ninja.exe"
```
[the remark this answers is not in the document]
hyp: Made them consistent, thanks for pointing that out.
I was setting `CMAKE_POSITION_INDEPENDENT_CODE` where we decided to set it in build.ps1, so the differences  stem from the original build.ps1 differences.

.github/workflows/swift-toolchain.yml
hyp: https://github.com/swiftlang/swift-installer-scripts/pull/308

.github/workflows/swift-toolchain.yml
hyp: Updated to use vswhere

.github/workflows/swift-toolchain.yml
hyp: I need these across both the Runtime, Foundation, XCTest and other base projects, so I'm not sure it makes sense to sink these flags into some cmake config as we will need one in each project

.github/workflows/swift-toolchain.yml
compnerd: I was thinking about the matrix entry itself. But I suppose that isn't too important.

compnerd (commented):
This feels very very close!

.github/workflows/swift-toolchain.yml
```diff
@@ -191,20 +208,25 @@ jobs:
...
-            echo CMAKE_SHARED_LINKER_FLAGS="" >> ${GITHUB_OUTPUT}
+            echo WINDOWS_CMAKE_C_FLAGS="/GS- /Gw /Gy /Oi /Oy /Zc:inline /Zc:preprocessor" >> ${GITHUB_OUTPUT}
+            echo WINDOWS_CMAKE_CXX_FLAGS="/GS- /Gw /Gy /Oi /Oy /Zc:inline /Zc:preprocessor /Zc:__cplusplus /D_DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR" >> ${GITHUB_OUTPUT}
+            echo WINDOWS_CMAKE_EXE_LINKER_FLAGS="" >> ${GITHUB_OUTPUT}
+            echo WINDOWS_CMAKE_SHARED_LINKER_FLAGS="" >> ${GITHUB_OUTPUT}
             echo CMAKE_Swift_FLAGS="-Xlinker -incremental:no -Xlinker -opt:ref -Xlinker -opt:icf" >> ${GITHUB_OUTPUT}
           fi
 
+          echo ANDROID_CMAKE_C_FLAGS="-ffunction-sections -fdata-sections -g" >> ${GITHUB_OUTPUT}
+          echo ANDROID_CMAKE_CXX_FLAGS="-ffunction-sections -fdata-sections -g" >> ${GITHUB_OUTPUT}
```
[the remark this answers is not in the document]
hyp: good idea!

.github/workflows/swift-toolchain.yml
```diff
@@ -239,7 +263,64 @@ jobs:
...
       fail-fast: false
       matrix:
-        arch: ['amd64', 'arm64', 'x86']
+        include:
+          - arch: amd64
+            cc: cl
+            cflags: ${{ needs.context.outputs.WINDOWS_CMAKE_C_FLAGS }}
+            cxx: cl
+            cxxflags: ${{ needs.context.outputs.WINDOWS_CMAKE_CXX_FLAGS }}
+            os: Windows
```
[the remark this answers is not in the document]
hyp: it's specifically to pass "Windows" to `CMAKE_SYSTEM_NAME`, and for general consistency with 'Android' too.

compnerd (approved):

PR #741, opened 2024-06-11
Pin ccache-action to a version that updates the PATH

kendalharland:
See https://github.com/hendrikmuhs/ccache-action/pull/204. This is necessary for this workflow to run on custom runners where `$USERPROFILE/.cargo/bin/` or any future installation path isn't already part of $GITHUB_PATH

### DIFF
--- a/.github/workflows/swift-toolchain.yml
+++ b/.github/workflows/swift-toolchain.yml
@@ -237,7 +237,10 @@ jobs:
           arch: ${{ matrix.arch }}
 
       - name: Setup sccache
-        uses: compnerd/ccache-action@sccache-0.7.4
+        # TODO: Revert to hendrikmuhs/ccache-action when https://github.com/hendrikmuhs/ccache-action/pull/204 is merged.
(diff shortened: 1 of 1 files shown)


tristanlabelle (approved):

compnerd:
I think that we should wait a bit and see if upstream responds. IIRC, they were pretty responsive, so giving them 24-48h seems reasonable.

kendalharland:
> I think that we should wait a bit and see if upstream responds. IIRC, they were pretty responsive, so giving them 24-48h seems reasonable.

I think this is reasonable. At the same time, waiting for upstream reviews from someone I can't directly contact, who also has their own priorities outside the project creates an undesirable slow down :/ I wonder if there is a less distributed strategy for managing changes to the 3p actions we depend on, such as keeping our own forks that are auto-synced, and retroactively upstreaming all changes when possible.

kendalharland:
@compnerd let's merge! Seems like upstream is not attending the PR at the moment

kendalharland:
Upstream has merged the change and this is no longer needed. I'll send a separate PR to revert back to hendrikmuhs/ccache-action

PR #742, opened 2024-06-13
Skip upload steps if the trigger is pull_request

kendalharland:
The workflow doesn't currently have a pull_request trigger (this decision is specific to the project that runs the workflow) but it's safe to assume that if a pull_request trigger is added, we should skip the snapshot and release steps. This PR adds if-statements to do so.


compnerd (approved):
Oh, this is a good idea!

PR #745, opened 2024-06-15
GHA: prefer to use a macro over changing the headers

compnerd:
`-D_ALLOW_COMPILER_AND_STL_VERSION_MISMATCH` should allow us to workaround the mismatch. This is cleaner than the solution of modifying system headers.


compnerd:
CC: @z2oh

z2oh:
Yeah that's much better. Not sure how I missed that `#ifndef`. Thanks!

z2oh (approved):

z2oh:
The broken invocation is actually of `swiftc`, so I think this needs to be passed to `CMAKE_Swift_FLAGS` as `-Xcc -D_ALLOW_COMPILER_AND_STL_VERSION_MISMATCH`. I'm testing this now, will file a PR soon.

PR #747, opened 2024-06-17
Upstream changes from thebrowsercompany/swift-build

kendalharland:
Pulls in some quality of life changes from thebrowsercompany/swift-build:
- Add an input to skip publishing artifacts.
- Pin ccache-action to a version that adds sccache to PATH after installing.
- Make sure Nuget sources are updated before installing python arm64
- Replicate swift_tag inpt for workflow_dispatch
- Don't publish stable.xml when artifact publishing is disabled.
- Don't create releases when artifact publishing is disabled.

compnerd/ccache-action can likely be removed now, as this workflow no longer depends on it and all changes have been upstreamed.

### DIFF
--- a/.github/workflows/swift-toolchain.yml
+++ b/.github/workflows/swift-toolchain.yml
@@ -7,6 +7,7 @@ on:
         description: 'Build Swift at a tagged snapshot'
         default: false
         type: boolean
+
       swift_tag:
         description: 'Swift Build Tag'
         required: false
@@ -21,16 +22,29 @@ on:
         description: 'Emit PDBs (Debug Info)'
         default: false
         type: boolean
+
       signed:
         description: 'Code Sign'
         default: false
         type: boolean
+
+      publish_artifacts:
+        description: 'If true, skip publishing release artifacts'
+        type: boolean
+        default: false
+        required: false
+
   workflow_call:
     inputs:
       snapshot:
         description: 'Build Swift at a tagged snapshot'
         default: false
         type: boolean
+
+      swift_tag:
+        description: 'Swift Build Tag'
+        required: false
+
       swift_version:
         description: 'Swift Version'
         default: '0.0.0'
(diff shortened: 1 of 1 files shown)


.github/workflows/swift-toolchain.yml
```diff
@@ -203,6 +247,7 @@ jobs:
           echo compilers_build_runner=${{ vars.COMPILERS_BUILD_RUNNER || vars.WINDOWS_BUILD_RUNNER || 'windows-latest' }} >> ${GITHUB_OUTPUT}
 
       - uses: actions/upload-artifact@v4
+        if: inputs.publish_artifacts != true
```
compnerd, line 250: This isn't really an artifact, it's merely tracking state for known valid builds and is what powers the `repo sync -s` to get a "stable" sync point.

.github/workflows/swift-toolchain.yml
```diff
@@ -21,16 +22,40 @@ on:
         description: 'Emit PDBs (Debug Info)'
         default: false
         type: boolean
+
       signed:
         description: 'Code Sign'
         default: false
         type: boolean
+
+      default_runner:
```
compnerd, line 31: I'm not convinced about the changes for the runners being an input. That's confusing for most people - it requires a good understanding of the labels which is not necessarily visible to everyone.

kendalharland: We can remove this for now then, and add it back once we think this workflow should expose an API for choosing the runners more generally.

.github/workflows/swift-toolchain.yml
kendalharland: Ah, that's right. This isn't actually necessary either, since we guard against `git push`ing this file in the snapshot job below. Thanks for catching. Removed.

.github/workflows/swift-toolchain.yml
```diff
@@ -239,7 +261,7 @@ jobs:
           arch: ${{ matrix.arch }}
 
       - name: Setup sccache
-        uses: compnerd/ccache-action@sccache-0.7.4
+        uses: hendrikmuhs/ccache-action@2e0e89e8d74340a03f75d58d02aae4c5ee1b15c6
```
compnerd, line 264: Would be nice to include what version this is.

compnerd (approved):

PR #754, opened 2024-06-27
Install the Android NDK before all steps that need it

kendalharland:
Ensure android NDK is installed regardless of runner.

 This fixes the workflow on runner images where the NDK is not
 installed. It bundles together a few other changes to ensure
 CMake is executed correctly and make it easier to diff CMake
 log in the GHA UI with the content in swift-toolchain.yml

## Changes

  * Always run gha-setup-vsdevenv even if runners.os != 'Windows'
    * It provides us with the path to CMake
  * Switch to nttdl/setup-ndk to install the Android NDK.
  * Always specify Android-specific MAKE defines, but enable others
    iff matrix.os == Android.
    * CMAKE_SYSTEM_NAME=${{ matrix.os }} controls which system we're
      building for, so there's no need to use matrix.extra_flags for
      Android-specific defines. Inlining them ensures powershell
      variable expansion works, and makes it easier to copy paste
      CMake commands for for local reproduction.

## Test

- BCNY Cirun: https://github.com/thebrowsercompany/swift-build/actions/runs/9717754718
- BCNY GitHub: https://github.com/thebrowsercompany/swift-build/actions/runs/9717754725

### DIFF
--- a/.github/workflows/swift-toolchain.yml
+++ b/.github/workflows/swift-toolchain.yml
@@ -295,7 +295,7 @@ jobs:
             cxx: clang++
             cxxflags: ${{ needs.context.outputs.ANDROID_CMAKE_CXX_FLAGS }}
             os: Android
-            extra_flags: -DCMAKE_ANDROID_API=${{ needs.context.outputs.ANDROID_API_LEVEL }} -DCMAKE_ANDROID_ARCH_ABI=arm64-v8a "-DCMAKE_MAKE_PROGRAM=$NINJA_PATH"
+            extra_flags: -DCMAKE_ANDROID_API=${{ needs.context.outputs.ANDROID_API_LEVEL }} -DCMAKE_ANDROID_ARCH_ABI=arm64-v8a
 
           - arch: armv7
             cc: clang
@@ -303,7 +303,7 @@ jobs:
             cxx: clang++
             cxxflags: ${{ needs.context.outputs.ANDROID_CMAKE_CXX_FLAGS }}
             os: Android
-            extra_flags: -DCMAKE_ANDROID_API=${{ needs.context.outputs.ANDROID_API_LEVEL }} -DCMAKE_ANDROID_ARCH_ABI=armeabi-v7a "-DCMAKE_MAKE_PROGRAM=$NINJA_PATH"
+            extra_flags: -DCMAKE_ANDROID_API=${{ needs.context.outputs.ANDROID_API_LEVEL }} -DCMAKE_ANDROID_ARCH_ABI=armeabi-v7a
 
           - arch: i686
             cc: clang
@@ -311,7 +311,7 @@ jobs:
             cxx: clang++
             cxxflags: ${{ needs.context.outputs.ANDROID_CMAKE_CXX_FLAGS }}
             os: Android
-            extra_flags: -DCMAKE_ANDROID_API=${{ needs.context.outputs.ANDROID_API_LEVEL }} -DCMAKE_ANDROID_ARCH_ABI=x86 "-DCMAKE_MAKE_PROGRAM=$NINJA_PATH"
+            extra_flags: -DCMAKE_ANDROID_API=${{ needs.context.outputs.ANDROID_API_LEVEL }} -DCMAKE_ANDROID_ARCH_ABI=x86
 
           - arch: x86_64
             cc: clang
@@ -319,7 +319,7 @@ jobs:
             cxx: clang++
             cxxflags: ${{ needs.context.outputs.ANDROID_CMAKE_CXX_FLAGS }}
             os: Android
-            extra_flags: -DCMAKE_ANDROID_API=${{ needs.context.outputs.ANDROID_API_LEVEL }} -DCMAKE_ANDROID_ARCH_ABI=x86_64 "-DCMAKE_MAKE_PROGRAM=$NINJA_PATH"
+            extra_flags: -DCMAKE_ANDROID_API=${{ needs.context.outputs.ANDROID_API_LEVEL }} -DCMAKE_ANDROID_ARCH_ABI=x86_64
 
     name: ${{ matrix.os }} ${{ matrix.arch }} SQLite3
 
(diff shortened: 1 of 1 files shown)


.github/workflows/swift-toolchain.yml
```diff
@@ -368,6 +373,8 @@ jobs:
                 -D CMAKE_MT=mt `
                 -D CMAKE_INSTALL_PREFIX=${{ github.workspace }}/BuildRoot/Library/sqlite-3.43.2/usr `
                 -D CMAKE_SYSTEM_NAME=${{ matrix.os }} `
+                -D CMAKE_ANDROID_NDK=$NDKPATH `
+                -D CMAKE_ANDROID_API=${{ needs.context.outputs.ANDROID_API_LEVEL }} `
```
compnerd, line 376: These are android specific, which can confuse the non-Android builds. Can we extract them somehow?

.github/workflows/swift-toolchain.yml
```diff
@@ -1119,6 +1135,9 @@ jobs:
                 -D CMAKE_MT=mt `
                 -D CMAKE_INSTALL_PREFIX=${{ github.workspace }}/BuildRoot/Library/zlib-1.3/usr `
                 -D CMAKE_SYSTEM_NAME=${{ matrix.os }} `
+                -D CMAKE_ANDROID_NDK=$NDKPATH `
+                -D CMAKE_ANDROID_API=${{ needs.context.outputs.ANDROID_API_LEVEL }} `
+                -D CMAKE_POSITION_INDEPENDENT_CODE=${{ matrix.os == 'Android' && 'YES' || 'NO' }} `
```
compnerd, line 1140: I think that we can set this unconditionally always, can't we?

.github/workflows/swift-toolchain.yml
```diff
@@ -1322,7 +1345,10 @@ jobs:
                 -D USE_WIN32_LARGE_FILES=${{ matrix.os == 'Windows' && 'YES' || 'NO' }} `
                 -D USE_WIN32_LDAP=NO `
                 -D ZLIB_ROOT=${{ github.workspace }}/BuildRoot/Library/zlib-1.3/usr  `
-                -D ZLIB_LIBRARY=${{ github.workspace }}/BuildRoot/Library/zlib-1.3/usr/lib/zlibstatic.lib
+                -D ZLIB_LIBRARY=${{ github.workspace }}/BuildRoot/Library/zlib-1.3/usr/lib/zlibstatic.lib `
+                -D CMAKE_POSITION_INDEPENDENT_CODE=${{ matrix.os == 'Android' && 'YES' || 'NO' }} `
```
compnerd, line 1349: I think that we can set this unconditionally always, can't we?

.github/workflows/swift-toolchain.yml
```diff
@@ -1446,7 +1476,10 @@ jobs:
                 -D LIBXML2_WITH_PYTHON=NO `
                 -D LIBXML2_WITH_TESTS=NO `
                 -D LIBXML2_WITH_THREADS=YES `
-                -D LIBXML2_WITH_ZLIB=NO
+                -D LIBXML2_WITH_ZLIB=NO `
+                -D CMAKE_POSITION_INDEPENDENT_CODE=${{ matrix.os == 'Android' && 'YES' || 'NO' }} `
```
compnerd, line 1480: I think that we can set this unconditionally always, can't we?

compnerd (commented):
This has multiple changes, including whitespace changes. Can you please split this up into focused changes?

.github/workflows/swift-toolchain.yml
kendalharland: Yes, actually. Thanks for pointing that out

.github/workflows/swift-toolchain.yml
kendalharland: If we set them conditionally (e.g. using `matrix.extra_flags`) then variables like $NDKPATH are not expanded and CMake configure fails. `CMAKE_ANDROID_API` does not suffer from this problem since it is not using powershell variables - We could move that one back to `matrix.extra_flags` but I left here for consistency. I could move that one back.

Re: confusing the build - IIUC these are ignored when `CMAKE_SYSTEM_NAME != 'Android'` based on the test builds in the PR description and on https://github.com/Kitware/CMake/blob/master/Modules/Platform/Android-Determine.cmake#L4.

.github/workflows/swift-toolchain.yml
kendalharland: Done.

.github/workflows/swift-toolchain.yml
kendalharland: Done.

.github/workflows/swift-toolchain.yml
compnerd: Yeah, but I'm still concerned because there are other weird behaviours with the Swift build and custom handling in the Swift modules which are not part of the CMake repository. I'd rather find a different way to handle that so that we can keep things stable.

kendalharland:
> This has multiple changes, including whitespace changes. Can you please split this up into focused changes?

Removed the whitespace changes.

Re the rest of the changes: Unfortunately these are all necessary to fix the Cirun builds :/ :
- Without unconditionally running gha-setup-vsdevenv, we can't find CMake
- Without inlining `CMAKE_ANDROID_NDK`, `$NDKPATH` is not expanded from `matrix.extra_flags` and CMake configure fails.
- Without switching to `nttdl/setup-ndk` I'll need to duplicate the powershell code to download the NDK before each cmake configure (Cirun images don't have the NDK)

compnerd:
I don't think I understand what you are suggesting. I don't see why you would not be able to extract just the installation of the NDK into a separate change that can go in first - that should have no impact on the GHA builds.

Separately, changing `gha-setup-devenv` could go in a separate change with no impact on GHA builders but further prepare the rules for supporting additional runners.

At that point, you could make just the changes needed to pass the extra require flags.

compnerd (approved):
This still conflates 2-3 different changes, but I suppose it is fine. The ninja changes are one item, the NDK installation and use are a separate item, and the PIC code is a third item. Please split changes into distinct issues being addressed in the future.

kendalharland:
I suppose that's true. Given how GitHub doesn't make it easy to stack PRs, it's just harder to test as 3 separate PRs. But you're correct, those individual changes would independently work on top of the existing GH runners. Will do in the future.

PR #756, opened 2024-07-01
Disable sccache when building ICU for Android

kendalharland:
These jobs are hitting a known fatal error in sccache where files aren't removed from the cache:

```
sccache: encountered fatal error
sccache: error: thread 'tokio-runtime-worker' panicked at src\lru_disk_cache\mod.rs:204:17: Error removing file from cache: `"C:\\a\\swift-build\\swift-build/.sccache\\preprocessor\\8\\f\\2\\8f2e7d1b9af7ff2cc23c6829fc3cae3044c631d88bf597901be5e04795779aad"`: failed to remove file `C:\a\swift-build\swift-build/.sccache\preprocessor\8\f\2\8f2e7d1b9af7ff2cc23c6829fc3cae3044c631d88bf597901be5e04795779aad`
sccache: caused by: thread 'tokio-runtime-worker' panicked at src\lru_disk_cache\mod.rs:204:17: Error removing file from cache: `"C:\\a\\swift-build\\swift-build/.sccache\\preprocessor\\8\\f\\2\\8f2e7d1b9af7ff2cc23c6829fc3cae3044c631d88bf597901be5e04795779aad"`: failed to remove file `C:\a\swift-build\swift-build/.sccache\preprocessor\8\f\2\8f2e7d1b9af7ff2cc23c6829fc3cae3044c631d88bf597901be5e04795779aad`
```

There's an open issue about this at https://github.com/mozilla/sccache/issues/2092.

## Test

https://github.com/thebrowsercompany/swift-build/actions/runs/9749014927

### DIFF
--- a/.github/workflows/swift-toolchain.yml
+++ b/.github/workflows/swift-toolchain.yml
@@ -569,6 +569,9 @@ jobs:
           arch: ${{ matrix.arch }}
 
       - name: Setup sccache
+        # TODO: Android ICU builds are consistently failing due to https://github.com/mozilla/sccache/issues/2092
+        # On GitHub and non-GitHub runner images. Reenable once this issue is resolved.
+        if: matrix.os != 'Android'
         uses: hendrikmuhs/ccache-action@2e0e89e8d74340a03f75d58d02aae4c5ee1b15c6
         with:
           max-size: 100M
@@ -590,10 +593,10 @@ jobs:
                 -D BUILD_DATA=${{ matrix.BUILD_DATA }} `
                 -D CMAKE_BUILD_TYPE=Release `
                 -D CMAKE_C_COMPILER=${{ matrix.cc }} `
-                -D CMAKE_C_COMPILER_LAUNCHER=sccache `
+                -D CMAKE_C_COMPILER_LAUNCHER=${{ matrix.os == 'Windows' && 'sccache' || '' }} `
                 -D CMAKE_C_FLAGS="${{ matrix.cflags }}" `
                 -D CMAKE_CXX_COMPILER=${{ matrix.cxx }} `
-                -D CMAKE_CXX_COMPILER_LAUNCHER=sccache `
+                -D CMAKE_CXX_COMPILER_LAUNCHER=${{ matrix.os == 'Windows' && 'sccache' || '' }} `
                 -D CMAKE_CXX_FLAGS="${{ matrix.cxxflags }}" `
                 -D CMAKE_MT=mt `
                 -D CMAKE_INSTALL_PREFIX=${{ github.workspace }}/BuildRoot/Library/icu-69.1/usr `


compnerd (changes requested):
Please inline the error message in the commit message so that we can follow what is going on, the logs are inaccessible due to the corporate security and broken Arc.

kendalharland:
> Please inline the error message in the commit message so that we can follow what is going on, the logs are inaccessible due to the corporate security and broken Arc.

Done

compnerd:
Hmm, its unclear if this is really the right approach - it just seems like you have a corrupted cache? What happens if you clear out the cache?

kendalharland:
> Hmm, its unclear if this is really the right approach - it just seems like you have a corrupted cache? What happens if you clear out the cache?

I don't think clearing the cache will fix things, given that this is flaking on both GitHub and Azure images where each Android ICU job executes in  a fresh VM instance. It seems like a bug in sccache that surfaces with higher probably when building for Android, so disabling it just unbreaks the build for the time being.

That being said, how does one clear the cache used by sccache? Where in the workflow could we do this if we wanted to workaround the error when building ICU?

compnerd:
You are missing the point of sccache ;-). The entire point is that the content *is* cached, even on a fresh VM image. You should look at the caches for the job, clearing out all the old runs should hopefully fix the issue.

kendalharland:
Ah, I see what you mean.  I just deleted each of the android ICU caches and will see if that works.

kendalharland:
That did it, thanks for the tip!

kendalharland:
Reopening because the sccache corruption comes back each time we attempt to target Android when building ICU, and it's not immediately clear what is causing this. We'll need to disable this while we debug.

hjyamauchi (approved):

kendalharland:
I'll need @compnerd to merge since I don't have write access

.github/workflows/swift-toolchain.yml
```diff
@@ -590,10 +593,10 @@ jobs:
                 -D BUILD_DATA=${{ matrix.BUILD_DATA }} `
                 -D CMAKE_BUILD_TYPE=Release `
                 -D CMAKE_C_COMPILER=${{ matrix.cc }} `
-                -D CMAKE_C_COMPILER_LAUNCHER=sccache `
+                -D CMAKE_C_COMPILER_LAUNCHER=${{ matrix.os == 'Windows' && 'sccache' || '' }} `
```
compnerd, line 596: Why is this change needed with the condition above?

compnerd (commented):
I do think that taking a closer look to the change is valuable - I'm concerned that this will remain disabled indefinitely.

kendalharland:
This is not needed after https://github.com/compnerd/swift-build/pull/757.  The reason for the cache misses is that the cache size was too small to begin with, so the first cache we create when building Android for ICU does not contain all of the expected entries causing sccache to subsequently fail when purging existing entries to make space

PR #757, opened 2024-07-09
Bump up the ICU build sccache size

hyp:
100Mib cache size is not able to fit all entries for the Android ICU build, which leads to manifestation of sccache bug happening in subsequent runs (https://github.com/mozilla/sccache/issues/2092).


hyp:
I tested this on two separate cache flows with multiple rounds producing full cache hits and no crashes.

compnerd (approved):
I wonder if we are building with debug information ... not that it is a bad thing.

PR #758, opened 2024-07-12
GHA: tweak workflow configuration

compnerd:
Reduce unnecessary verbosity on the configuration. Always generate the snapshot as we want to track any successful build. Add some whitespace to the release creation script.

### DIFF
--- a/.github/workflows/swift-toolchain.yml
+++ b/.github/workflows/swift-toolchain.yml
@@ -29,8 +29,8 @@ on:
         default: false
         type: boolean
 
-      publish_artifacts:
-        description: 'If true, publish release artifacts'
+      create_release:
+        description: 'Create Release'
         type: boolean
         default: true
         required: false
@@ -69,8 +69,8 @@ on:
         default: false
         type: boolean
 
-      publish_artifacts:
-        description: 'If true, publish release artifacts'
+      create_release:
+        description: 'Create Release'
         type: boolean
         default: true
         required: false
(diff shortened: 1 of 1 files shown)


compnerd:
CC: @kendalharland

.github/workflows/swift-toolchain.yml
```diff
@@ -3332,7 +3332,7 @@ jobs:
   snapshot:
     runs-on: ubuntu-latest
     needs: [context, smoke_test]
-    if: inputs.publish_artifacts == true
+    if: github.event_name == 'schedule'
```
kendalharland, line 3335: I think this one should stay as `inputs.publish_artifacts`, primarily for the cases where we want to produce a one-off build and the event is `workflow_dispatch`.  Setting it to `github.event_name == 'schedule'` is BNCY-specific (or at least that was my understanding) WDYT?

kendalharland (approved):

.github/workflows/swift-toolchain.yml
compnerd: Hmm, we could make `create_release` true by default, but we should still update the stable point if the build is known to succeed at that point.

compnerd: Discussed offline; this is the simplest solution to ensure that the snapshot is monotonically increasing. While one-offs would be nice to use for snapshotting purposes, we do not know if they are triggered at an older point and thus could be pushing the stable point backwards instead of forwards.

kendalharland: resolved offline: basically what you said above. Using `schedule` as you have here and ignoring the rare case where manual builds don't update stable.xml is the simplest thing to do 👍🏾

PR #759, opened 2024-07-12
GHA: tweak dispatch workflow inputs further

compnerd:
This reorders the options to be inputs first, makes the Android API level lose the asterisk as there is a default value, and infer the snapshot building based on the input.


compnerd:
CC: @kendalharland

kendalharland (approved):

PR #762, opened 2024-07-15
 Install VS build tools 14.39.33519 for sdk and devtools steps on release/5.10

kendalharland:
The SDK and devtools steps on the release/5.10 branch must also use the old version `14.39.33519` of VS build
tools which are still compatible with Clang 16.  

## Test

https://github.com/thebrowsercompany/swift-build/actions/runs/9943835623


compnerd (approved):
Gross!

PR #764, opened 2024-07-16
Update the depedency tags.

hjyamauchi:
Based on https://github.com/swiftlang/swift/blob/7eff8fbb44007d60b998eec33e128e30056dc9db/utils/update_checkout/update-checkout-config.json#L114


compnerd (approved):
Updates are good!

PR #770, opened 2024-07-24
Override cmake to avoid the cmake 3.30 issue.

hjyamauchi:


.github/workflows/swift-toolchain.yml
```diff
@@ -1686,6 +1686,11 @@ jobs:
           release-asset-name: installer-amd64.exe
           release-tag-name: '20231016.0'
 
+      # workaround CMake 3.30 issue
+      - uses: jwlawson/actions-setup-cmake@v2
+        with:
+          cmake-version: '3.29.3'
```
compnerd, line 1692:
```suggestion
          cmake-version: '3.29'
```

PR #777, opened 2024-07-29
Update android corelibs-foundation pinned checkout

hyp:
The checkout needs to be updated after the posix_filesystem apinote fix landed in Swift. I needed to switch to my own remote as I could no longer push a new branch into the Apple remote.

### DIFF
--- a/.github/workflows/swift-toolchain.yml
+++ b/.github/workflows/swift-toolchain.yml
@@ -3097,7 +3097,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           repository: apple/swift-installer-scripts
-          ref: e1adc0d5dde137fa5941316e89a3ed139a273dde
+          ref: 633843b00fc8024dd86674fe43373ae4fd0f2229
           path: ${{ github.workspace }}/SourceCache/swift-installer-scripts
           show-progress: false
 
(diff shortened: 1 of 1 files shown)


hyp:
waiting to validate first in a run before merging

.github/workflows/swift-toolchain.yml
```diff
@@ -1653,8 +1653,8 @@ jobs:
       - uses: actions/checkout@v4
         if: matrix.os == 'Android'
         with:
-          repository: apple/swift-corelibs-foundation
-          ref: a515f3e15c81d029411c540df7edf52d427ff2cb
+          repository: hyp/swift-corelibs-foundation
+          ref: 6bf677693eb23030eb3646c9b121c665f40aa8b0
```
compnerd, line 1657: Is this for the changes wrt modulemaps?

hyp: it's dropping the `fts_open` changes we had for Android after the posix_fielsystem api note fix landed (https://github.com/swiftlang/swift/pull/75494), since now Swift applies the api note to the `fts_open` API correctly in the workflow.

hjyamauchi (approved):

.github/workflows/swift-toolchain.yml
compnerd: I wonder if we can get the changes in upstream. I suppose that this is fine for now.

compnerd (approved):

.github/workflows/swift-toolchain.yml
hyp: Yeah this will all go away once recore is enabled in the workflow.

PR #783, opened 2024-08-01
workaround smoke-test t-s-c bug by dropping SDKROOT slash

hyp:


hjyamauchi (approved):

compnerd (changes requested):
Actually, I'm not sure that we want to do this. This is the value that is configured and while this will get us a build, it will break for every user.

PR #786, opened 2024-08-06
GHA: remove ICU build after the re-core transition

compnerd:
Remove the ICU dependency as it has been removed from the repo manifest. This now requires that we use the re-cored Foundation.


compnerd:
CC: @steelskin

compnerd:
This does require the recore being enabled for Android unfortunately.

hyp (approved):

PR #789, opened 2024-08-12
GHA: build Foundation Macros

compnerd:
Upstream now builds and packages the swift macros unconditionally. Update the build rules to do the same.


hyp (approved):
Thanks!

PR #794, opened 2024-08-13
GHA: split out standard library phase

compnerd:
Split out the standard library build phase to allow us to build the macros for the correct hosts. This is required to allow us to build the macros for distribution and for the build of the SDK.


hyp (approved):
That should work all the way up to foundation for Android (as it's still waiting on LP32 fix to land)!

PR #795, opened 2024-08-14
GHA: use `gh` to create release

compnerd:
Migrate from actions to `gh` to create the release to avoid some warnings. This also ensures that we no longer accidentally mark the 5.10 releases as the latest.


hyp (approved):
This already works on other branches right so should be good

PR #796, opened 2024-08-14
[sqlite] Remove Android builds

Steelskin:
These are not consumed by any other job or step.
In addition, this parameterizes the SQLite version, simplifying future updates.


compnerd (approved):
Ah, right sqlite3 is used by SPM and llbuild, not by the SDK!

PR #793, opened 2024-08-13
[Mac build] Move build to sub-workflow

Steelskin:
This is a preliminary step to eventually add the Mac build. The Mac build will be using a separate job with its own matrix.

PR #797, opened 2024-08-14
[Mac build] Rename builder to a generic name

Steelskin:
`windows_build_runner` is too Windows-specific. This renames the variable to `default_build_runner`, to make space for future Darwin builds.

The variable is kept to its original name in the `context` step, since this step will eventually also output Darwin variables.

### DIFF
--- a/.github/workflows/swift-toolchain.yml
+++ b/.github/workflows/swift-toolchain.yml
@@ -119,10 +119,10 @@ jobs:
       yams_revision: ${{ steps.context.outputs.yams_revision }}
       zlib_revision: ${{ steps.context.outputs.zlib_revision }}
       ANDROID_API_LEVEL: ${{ steps.context.outputs.ANDROID_API_LEVEL }}
-      HOST_CMAKE_C_FLAGS: ${{ steps.context.outputs.WINDOWS_CMAKE_C_FLAGS }}
-      HOST_CMAKE_CXX_FLAGS: ${{ steps.context.outputs.WINDOWS_CMAKE_CXX_FLAGS }}
-      HOST_CMAKE_EXE_LINKER_FLAGS: ${{ steps.context.outputs.WINDOWS_CMAKE_EXE_LINKER_FLAGS }}
-      HOST_CMAKE_SHARED_LINKER_FLAGS: ${{ steps.context.outputs.WINDOWS_CMAKE_SHARED_LINKER_FLAGS }}
+      WINDOWS_CMAKE_C_FLAGS: ${{ steps.context.outputs.WINDOWS_CMAKE_C_FLAGS }}
+      WINDOWS_CMAKE_CXX_FLAGS: ${{ steps.context.outputs.WINDOWS_CMAKE_CXX_FLAGS }}
+      WINDOWS_CMAKE_EXE_LINKER_FLAGS: ${{ steps.context.outputs.WINDOWS_CMAKE_EXE_LINKER_FLAGS }}
+      WINDOWS_CMAKE_SHARED_LINKER_FLAGS: ${{ steps.context.outputs.WINDOWS_CMAKE_SHARED_LINKER_FLAGS }}
       ANDROID_CMAKE_C_FLAGS: ${{ steps.context.outputs.ANDROID_CMAKE_C_FLAGS }}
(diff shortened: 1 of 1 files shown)


.github/workflows/swift-toolchain.yml
```diff
@@ -119,10 +119,10 @@ jobs:
...
       zlib_revision: ${{ steps.context.outputs.zlib_revision }}
       ANDROID_API_LEVEL: ${{ steps.context.outputs.ANDROID_API_LEVEL }}
-      WINDOWS_CMAKE_C_FLAGS: ${{ steps.context.outputs.WINDOWS_CMAKE_C_FLAGS }}
-      WINDOWS_CMAKE_CXX_FLAGS: ${{ steps.context.outputs.WINDOWS_CMAKE_CXX_FLAGS }}
-      WINDOWS_CMAKE_EXE_LINKER_FLAGS: ${{ steps.context.outputs.WINDOWS_CMAKE_EXE_LINKER_FLAGS }}
-      WINDOWS_CMAKE_SHARED_LINKER_FLAGS: ${{ steps.context.outputs.WINDOWS_CMAKE_SHARED_LINKER_FLAGS }}
+      HOST_CMAKE_C_FLAGS: ${{ steps.context.outputs.WINDOWS_CMAKE_C_FLAGS }}
+      HOST_CMAKE_CXX_FLAGS: ${{ steps.context.outputs.WINDOWS_CMAKE_CXX_FLAGS }}
+      HOST_CMAKE_EXE_LINKER_FLAGS: ${{ steps.context.outputs.WINDOWS_CMAKE_EXE_LINKER_FLAGS }}
+      HOST_CMAKE_SHARED_LINKER_FLAGS: ${{ steps.context.outputs.WINDOWS_CMAKE_SHARED_LINKER_FLAGS }}
```
compnerd, line 125: These really should be Windows - they contain Windows specific flags, which is why we have the `ANDROID_CMAKE_C_FLAGS` as well. If you want to collapse these somehow then we should only have `CMAKE_C_FLAGS`. We also have more than one host that we build for, so `HOST_` is a poor choice of prefix.

.github/workflows/swift-toolchain.yml
```diff
@@ -2620,7 +2620,7 @@ jobs:
   debugging_tools:
     name: Debugging Tools
     needs: [context, compilers, devtools, stdlib, sdk]
-    runs-on: ${{ needs.context.outputs.windows_build_runner }}
+    runs-on: ${{ needs.context.outputs.libs_build_runner }}
```
compnerd, line 2623: This renaming doesn't make sense - the debug tools are not libraries.

.github/workflows/swift-toolchain.yml
Steelskin: This will make a bit more sense once #793 has landed but it is stalled on fixing the build. The idea here is that these are going to be referring to flags used for the host platform (either Mac or Windows). When sub-workflow is set up, `HOST_CMAKE_*_FLAGS` will be set by the main workflow to either `WINDOWS_CMAKE_*_FLAGS` or `DARWIN_CMAKE_*_FLAGS`, depending on the build.

.github/workflows/swift-toolchain.yml
Steelskin: Renamed to `main_build_runner`. Same as the other comment, this will make more sense once #793 has landed.

.github/workflows/swift-toolchain.yml
compnerd: The problem is that there is no one canonical host. We have the host for the toolchain and the host for the runtimes which are different. Technically, it would be host and target, but we don't call the target flags the target flags but name them according to the target.

.github/workflows/swift-toolchain.yml
compnerd: The renaming here is generally reasonable, as `windows_build_runner` isn't great. I think that `default_build_runner` is a better name though.

.github/workflows/swift-toolchain.yml
Steelskin: Ack. I am abandoning that part of the change and will make a separate change to split the usage.

.github/workflows/swift-toolchain.yml
Steelskin: Renamed.

compnerd (approved):

compnerd (approved):

PR #800, opened 2024-08-15
GHA: tweak some names (NFC)

compnerd:
Clean up the job names to be consistent throughout.

### DIFF
--- a/.github/workflows/swift-toolchain.yml
+++ b/.github/workflows/swift-toolchain.yml
@@ -291,6 +291,8 @@ jobs:
       matrix:
         arch: ['amd64', 'arm64']
 
+    name: Windows ${{ matrix.arch }} CMark GFM
+
     steps:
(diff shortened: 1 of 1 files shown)


Steelskin (approved):

.github/workflows/swift-toolchain.yml
```diff
@@ -342,6 +344,8 @@ jobs:
     needs: [cmark_gfm]
     runs-on: ${{ inputs.default_build_runner }}
 
+    name: Windows amd64 Compiler Build Tools
```
hjyamauchi, line 347: Could this be called "build-arch" instead of "amd64" even though it's hardcoded at this point?

compnerd: I'd love to make that a matrix field, but right now, we hardcode build to amd64 as GH does not provide non-x64 build hosts.

PR #803, opened 2024-08-16
GHA: repair the Windows runtime/SDK packaging

compnerd:
We now need to download the standard library and the SDK to get the same content that we used to previously have in just the SDK. Adjust that accordingly.


Steelskin (approved):

compnerd:
This gets us to the next failure (missing swift-tesitng).

PR #806, opened 2024-08-17
[Mac build] Add optional Mac build

Steelskin:
This adds a new Mac workflow, currently disabled by default and adapts the sqlite job to build on Mac. All other jobs are currently only enabled for Windows, they will be adapted in subsequent PRs.

### DIFF
--- a/.github/workflows/build-toolchain.yml
+++ b/.github/workflows/build-toolchain.yml
@@ -134,6 +134,8 @@ jobs:
       WINDOWS_CMAKE_CXX_FLAGS: ${{ steps.context.outputs.WINDOWS_CMAKE_CXX_FLAGS }}
       WINDOWS_CMAKE_EXE_LINKER_FLAGS: ${{ steps.context.outputs.WINDOWS_CMAKE_EXE_LINKER_FLAGS }}
       WINDOWS_CMAKE_SHARED_LINKER_FLAGS: ${{ steps.context.outputs.WINDOWS_CMAKE_SHARED_LINKER_FLAGS }}
+      DARWIN_CMAKE_C_FLAGS: ${{ steps.context.outputs.DARWIN_CMAKE_C_FLAGS }}
+      DARWIN_CMAKE_CXX_FLAGS: ${{ steps.context.outputs.DARWIN_CMAKE_CXX_FLAGS }}
       ANDROID_CMAKE_C_FLAGS: ${{ steps.context.outputs.ANDROID_CMAKE_C_FLAGS }}
       ANDROID_CMAKE_CXX_FLAGS: ${{ steps.context.outputs.ANDROID_CMAKE_CXX_FLAGS }}
       ANDROID_CMAKE_EXE_LINKER_FLAGS: ${{ steps.context.outputs.ANDROID_CMAKE_EXE_LINKER_FLAGS }}
@@ -144,7 +146,10 @@ jobs:
       swift_version: ${{ steps.context.outputs.swift_version }}
       swift_tag: ${{ steps.context.outputs.swift_tag }}
       windows_build_runner: ${{ steps.context.outputs.windows_build_runner }}
-      compilers_build_runner: ${{ steps.context.outputs.compilers_build_runner }}
+      windows_compilers_runner: ${{ steps.context.outputs.windows_compilers_runner }}
+      mac_build_runner: ${{ steps.context.outputs.mac_build_runner }}
+      windows_host_matrix: ${{ steps.setup-matrix.outputs.windows_host_matrix }}
+      darwin_host_matrix: ${{ steps.setup-matrix.outputs.darwin_host_matrix }}
     steps:
       - id: context
         name: Generate Build Context
@@ -219,6 +224,8 @@ jobs:
             echo debug_info=true >> ${GITHUB_OUTPUT}
             echo WINDOWS_CMAKE_C_FLAGS="/GS- /Gw /Gy /Oi /Oy /Z7 /Zc:inline /Zc:preprocessor" >> ${GITHUB_OUTPUT}
             echo WINDOWS_CMAKE_CXX_FLAGS="/GS- /Gw /Gy /Oi /Oy /Z7 /Zc:inline /Zc:preprocessor /Zc:__cplusplus /D_DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR" >> ${GITHUB_OUTPUT}
+            echo DARWIN_CMAKE_C_FLAGS="-g" >> ${GITHUB_OUTPUT}
+            echo DARWIN_CMAKE_CXX_FLAGS="-g" >> ${GITHUB_OUTPUT}
             echo ANDROID_CMAKE_C_FLAGS="-ffunction-sections -fdata-sections -g" >> ${GITHUB_OUTPUT}
             echo ANDROID_CMAKE_CXX_FLAGS="-ffunction-sections -fdata-sections -g" >> ${GITHUB_OUTPUT}
             echo WINDOWS_CMAKE_EXE_LINKER_FLAGS="-incremental:no -debug -opt:ref -opt:icf" >> ${GITHUB_OUTPUT}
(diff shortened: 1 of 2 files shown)


hjyamauchi:
Is `build-toolchain.yml` for mac and `swift-toolchain.yml` for windows?

.github/workflows/build-toolchain.yml
```diff
@@ -228,6 +233,8 @@ jobs:
             echo debug_info=true >> ${GITHUB_OUTPUT}
             echo WINDOWS_CMAKE_C_FLAGS="/GS- /Gw /Gy /Oi /Oy /Z7 /Zc:inline /Zc:preprocessor" >> ${GITHUB_OUTPUT}
             echo WINDOWS_CMAKE_CXX_FLAGS="/GS- /Gw /Gy /Oi /Oy /Z7 /Zc:inline /Zc:preprocessor /Zc:__cplusplus /D_DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR" >> ${GITHUB_OUTPUT}
+            echo DARWIN_CMAKE_C_FLAGS="-g" >> ${GITHUB_OUTPUT}
+            echo DARWIN_CMAKE_CXX_FLAGS="-g" >> ${GITHUB_OUTPUT}
```
compnerd, line 237: I wonder if we should use `RelWithDebInfo` instead of adding `-g`.

.github/workflows/build-toolchain.yml
```diff
@@ -275,11 +286,62 @@ jobs:
...
+          echo "windows_host_matrix=$(jq -r -c '.' <<< ${WINDOWS_HOST_MATRIX})" >> ${GITHUB_OUTPUT}
+          echo "darwin_host_matrix=$(jq -r -c '.' <<< ${DARWIN_HOST_MATRIX})" >> ${GITHUB_OUTPUT}
+
   windows-build:
     needs: [context]
     name: Windows Swift Toolchains Build
     uses: ./.github/workflows/swift-toolchain.yml
     with:
+      builder_os: Windows
+      builder_arch: amd64
```
compnerd, line 344: What is "builder"? Is this the build or the host?

.github/workflows/build-toolchain.yml
```diff
@@ -313,6 +375,8 @@ jobs:
       yams_revision: ${{ needs.context.outputs.yams_revision }}
       zlib_revision: ${{ needs.context.outputs.zlib_revision }}
       ANDROID_API_LEVEL: ${{ needs.context.outputs.ANDROID_API_LEVEL }}
+      HOST_CMAKE_C_FLAGS: ${{ needs.context.outputs.WINDOWS_CMAKE_C_FLAGS }}
+      HOST_CMAKE_CXX_FLAGS: ${{ needs.context.outputs.WINDOWS_CMAKE_CXX_FLAGS }}
```
compnerd, line 379: What is HOST here? Is it the host or the build? CMake mixes up the terms, and that makes this very confusing. I'd prefer a different name unless this is something that is baked into CMake.

.github/workflows/swift-toolchain.yml
```diff
@@ -3,6 +3,18 @@ name: Swift Toolchain Build
 on:
   workflow_call:
     inputs:
+      builder_os:
+        required: true
+        type: string
+
+      builder_arch:
+        required: true
+        type: string
```
compnerd, line 12: Please use standard terms, `builder` is not one of them: `build`, `host`, `target`.

.github/workflows/swift-toolchain.yml
```diff
@@ -135,6 +147,14 @@ on:
...
         type: string
 
+      HOST_CMAKE_C_FLAGS:
+        required: true
+        type: string
+
+      HOST_CMAKE_CXX_FLAGS:
+        required: true
+        type: string
+
```
compnerd, line 157: If these are for the host, I'd prefer `CMAKE_C_FLAGS` and `CMAKE_CXX_FLAGS`.

.github/workflows/swift-toolchain.yml
```diff
@@ -205,31 +224,17 @@ on:
         required: true
 
 env:
-  SCCACHE_DIRECT: yes
+  SCCACHE_DIRECT: on
```
compnerd, line 227: Why this change?

.github/workflows/swift-toolchain.yml
```diff
@@ -245,11 +250,20 @@ jobs:
           show-progress: false
 
       - uses: compnerd/gha-setup-vsdevenv@main
+        if: inputs.builder_os == 'Windows'
```
compnerd, line 253: I think that this is a no-op on other targets, so we can safely just run it always.

.github/workflows/swift-toolchain.yml
```diff
@@ -245,11 +250,20 @@ jobs:
...
         with:
-          host_arch: amd64
+          host_arch: ${{ inputs.builder_arch }}
           components: 'Microsoft.VisualStudio.Component.VC.Tools.x86.x64;Microsoft.VisualStudio.Component.VC.Tools.ARM64'
           arch: ${{ matrix.arch }}
 
+      - uses: maxim-lobanov/setup-xcode@v1
```
compnerd, line 259: This is not a whitelisted action, I think this will fail. Do we need to add this dependency?

.github/workflows/swift-toolchain.yml
```diff
@@ -245,11 +250,20 @@ jobs:
...
+        if: inputs.builder_os == 'Darwin'
+        with:
+          xcode-version: latest-stable
+
+      - uses: seanmiddleditch/gha-setup-ninja@master
```
compnerd, line 264: Likewise.

Steelskin:
> Is `build-toolchain.yml` for mac and `swift-toolchain.yml` for windows?

No, `build-toolchain.yml` is the "entry point" and it calls `swift-toolchain.yml`, which performs most of the build tasks. This change adds a second `swift-toolchain.yml` workflow, with Mac as the runner to build Mac toolchains.

.github/workflows/swift-toolchain.yml
compnerd: Don't create a new action - this one is fine to use. Just that before we can merge this, we will need to configure the permissions.

.github/workflows/swift-toolchain.yml
compnerd: Okay, we can use the action, but will need to configure the repo before merging.

compnerd (approved):

.github/workflows/build-toolchain.yml
```diff
@@ -256,7 +265,9 @@ jobs:
           fi
 
           echo windows_build_runner=${{ inputs.windows_default_runner || vars.WINDOWS_BUILD_RUNNER || 'windows-latest' }} >> ${GITHUB_OUTPUT}
-          echo compilers_build_runner=${{ inputs.windows_compilers_runner || inputs.windows_default_runner || vars.COMPILERS_BUILD_RUNNER || vars.WINDOWS_BUILD_RUNNER || 'windows-latest' }} >> ${GITHUB_OUTPUT}
+          echo windows_compilers_runner=${{ inputs.windows_compilers_runner || inputs.windows_default_runner || vars.COMPILERS_BUILD_RUNNER || vars.WINDOWS_BUILD_RUNNER || 'windows-latest' }} >> ${GITHUB_OUTPUT}
```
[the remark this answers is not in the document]
Steelskin: It's one big "if else if", it might be more readable written as such. We can follow-up.

.github/workflows/swift-toolchain.yml
Steelskin: It worked when I tried it because the downstream repo is configured to accept all actions. We should change that. I don't have access to this repo configuration so I don't know if it works here.

compnerd: Oh, if downstream has that ... let it fly. I'm less concerned about that restriction on upstream and more on downstream.

.github/workflows/build-toolchain.yml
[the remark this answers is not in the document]
compnerd: c'est bon

.github/workflows/swift-toolchain.yml
Steelskin: I restored the ninja action.

PR #809, opened 2024-08-20
GHA: wire up Foundation macros to swift-corelibs-foundation

compnerd:
We are already building the Foundation macros but did not pass that along to the Foundation build. This is now required and failure to pass it along results in a build failure. Wire it up so that we can build and avoid another build of the macros.


compnerd:
Going to merge to just try to brute force the CI.

PR #811, opened 2024-08-20
Add the workspace hash to ccache key

Steelskin:
Depending on the runner configuration, `github.workspace` might be set up differently, resulting in issues when restoring previous builds from sccache. In order to work around this issue, this adds a hash of `github.workspace` to the sccache key to uniquely identify the build between different runner types.


compnerd (commented):
I think that we should be looking at the runner context rather than the inputs context. https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#runner-context

compnerd (approved):

PR #812, opened 2024-08-21
Use runners from vars if available

Steelskin:
This was removed in #801 to bring the file in line with downstream. However, it is better to rely on the vars environment when invoking the main workflow from dispatch, while offering an override when invoking the workflow as a sub-workflow.


Steelskin:
@compnerd We should make a decision on #811 first or this is going to break downstream.

compnerd (approved):

PR #815, opened 2024-08-23
fix devtools stage

hyp:
Fix the build failures in the devtools stage

### DIFF
--- a/.github/workflows/swift-toolchain.yml
+++ b/.github/workflows/swift-toolchain.yml
@@ -2416,7 +2416,7 @@ jobs:
                 -D CMAKE_Swift_COMPILER=${SWIFTC} `
                 -D CMAKE_Swift_COMPILER_TARGET=${{ matrix.triple }} `
                 -D CMAKE_Swift_COMPILER_WORKS=YES `
-                -D CMAKE_Swift_FLAGS="${{ inputs.CMAKE_Swift_FLAGS }}" `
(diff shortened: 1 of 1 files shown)


.github/workflows/swift-toolchain.yml
```diff
@@ -2416,7 +2410,7 @@ jobs:
                 -D CMAKE_Swift_COMPILER=${SWIFTC} `
                 -D CMAKE_Swift_COMPILER_TARGET=${{ matrix.triple }} `
                 -D CMAKE_Swift_COMPILER_WORKS=YES `
-                -D CMAKE_Swift_FLAGS="${{ inputs.CMAKE_Swift_FLAGS }}" `
```
Steelskin, line 2419: Should this be appended rather than replaced?
(Same for all other instances of this replacement)

hyp: appended where?

compnerd (approved):
Ugh, is really unfortunate that just picking it up from the environment doesn't work equivalently.

.github/workflows/swift-toolchain.yml
Steelskin: Oh nevermind, I was on mobile and misread the diff.

Steelskin (approved):

hjyamauchi (approved):

Steelskin (approved):

PR #816, opened 2024-08-26
[Mac build] Build cmark-gfm on Mac

Steelskin:


.github/workflows/swift-toolchain.yml
```diff
@@ -304,16 +308,17 @@ jobs:
...
-      matrix:
-        arch: ['amd64', 'arm64']
+      matrix: ${{ fromJSON(inputs.host_matrix) }}
 
-    name: Windows ${{ matrix.arch }} CMark GFM
+    name: ${{ matrix.os }} ${{ matrix.arch }} CMark GFM
+
+    defaults:
+      run:
+        shell: pwsh
```
compnerd, line 321: Can we do defaults on the workflow rather than the jobs?

.github/workflows/swift-toolchain.yml
```diff
@@ -325,13 +330,15 @@ jobs:
 
       - uses: compnerd/gha-setup-vsdevenv@main
         with:
-          host_arch: amd64
+          host_arch: ${{ inputs.build_arch }}
           components: 'Microsoft.VisualStudio.Component.VC.Tools.x86.x64;Microsoft.VisualStudio.Component.VC.Tools.ARM64'
           arch: ${{ matrix.arch }}
 
+      - uses: seanmiddleditch/gha-setup-ninja@master
+        if: inputs.build_os == 'Darwin'
```
compnerd, line 338: If we are doing this, can we perhaps file an issue for a follow up which uses setup-cmake and gha-setup-ninja to configure those rather than the vsdevenv on Windows?

Steelskin: Filed https://github.com/compnerd/swift-build/issues/821

.github/workflows/swift-toolchain.yml
Steelskin: Done.

compnerd (approved):

PR #817, opened 2024-08-27
[Macros] Remove upload exe step

Steelskin:
The "macros" jobs are configured to only build shared libraries.


compnerd (approved):

compnerd (commented):
Please correct the commit message before submitting - macro builds *do* generate exes, however, we are currently building these as shared libraries.

PR #818, opened 2024-08-27
Set boolean variables to boolean types

Steelskin:
These were previously set to string types, resulting in incorrect
evaluation in `if` blocks.


compnerd (changes requested):
Please change the type of the input itself. They are currently incorrectly marked as `string` rather than `boolean`.

Steelskin:
> Please change the type of the input itself. They are currently incorrectly marked as `string` rather than `boolean`.

This is done, I had to modify the way we pass the variable but it works now (verified on downstream CI)

hjyamauchi (approved):

compnerd (approved):

PR #822, opened 2024-08-28
[Mac build] Add build_tools to the Mac build

Steelskin:


.github/workflows/swift-toolchain.yml
```diff
@@ -485,17 +497,17 @@ jobs:
...
-            ${{ github.workspace }}/BinaryCache/0/bin/swift-compatibility-symbols.exe
+            ${{ github.workspace }}/BinaryCache/0/bin/llvm-tblgen${{ inputs.exe_suffix }}
+            ${{ github.workspace }}/BinaryCache/0/bin/clang-tblgen${{ inputs.exe_suffix }}
+            ${{ github.workspace }}/BinaryCache/0/bin/lldb-tblgen${{ inputs.exe_suffix }}
+            ${{ github.workspace }}/BinaryCache/0/bin/llvm-config${{ inputs.exe_suffix }}
+            ${{ github.workspace }}/BinaryCache/0/bin/clang-pseudo-gen${{ inputs.exe_suffix }}
+            ${{ github.workspace }}/BinaryCache/0/bin/clang-tidy-confusable-chars-gen${{ inputs.exe_suffix }}
+            ${{ github.workspace }}/BinaryCache/0/bin/swift-def-to-strings-converter${{ inputs.exe_suffix }}
+            ${{ github.workspace }}/BinaryCache/0/bin/swift-serialize-diagnostics${{ inputs.exe_suffix }}
+            ${{ github.workspace }}/BinaryCache/0/bin/swift-compatibility-symbols${{ inputs.exe_suffix }}
```
compnerd, line 510: Can we not use a pattern instead of the exe suffix?

Steelskin: What do you mean by "using a pattern"?

hjyamauchi: I'm guessing regex-like path pattern matching? I don't know if this works but maybe something like `{{ github.workspace }}/BinaryCache/0/bin/llvm-tblgen(.exe)` (where `.exe` may or may not exist) might work? This may be relevant: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

Steelskin: This action uses this particular glob library: https://github.com/actions/toolkit/blob/main/packages/glob/README.md

Unfortunately, it does not support `(optional-pattern)` and only supports `?` as "any single character", not as "0 or 1 of the previous character", so neither `(.exe)` nor `.?e?x?e?` worked. I am open to other suggestions.

hjyamauchi: I wonder if `.*` as in `../llvm-tblgen.*` would be. I think this should be fine as long as there isn't a path that is a subpart of another (likely the case here)? If this doesn't pan out, I suggest sticking to what you had originally.

Steelskin: I thought about it but my concern is over-packaging. We may end up with a bunch of large unrelated files like PDBs.

compnerd: Ah, that's fair. Unfortunate that we cannot simply glob.

hjyamauchi: A good point about the pdb and such

Steelskin: I moved to exporting the names as a sub-step. We'll have to change the consumption targets down the line but this can wait when I get there.

.github/workflows/swift-toolchain.yml
```diff
@@ -483,19 +491,37 @@ jobs:
...
         run: cmake --build ${{ github.workspace }}/BinaryCache/0 --target swift-compatibility-symbols
 
+      - name: Export binary paths
+        id: export-binary-paths
+        run: |
+          if ( "${{ matrix.os }}" -eq "Windows" ) {
+            $Suffix = ".exe"
+          } else {
+            $Suffix = ""
+          }
```
compnerd, line 501:
```suggestion
          $Suffix = if ( "${{ matrix.os }}" -eq "Windows" ) { ".exe" } else { "" }
```

.github/workflows/swift-toolchain.yml
```diff
@@ -483,19 +491,33 @@ jobs:
...
+          echo "llvm_tblgen=${{ github.workspace }}/BinaryCache/0/bin/llvm-tblgen${Suffix}" >> $env:GITHUB_OUTPUT
+          echo "clang_tblgen=${{ github.workspace }}/BinaryCache/0/bin/clang-tblgen${Suffix}" >> $env:GITHUB_OUTPUT
+          echo "lldb_tblgen=${{ github.workspace }}/BinaryCache/0/bin/lldb-tblgen${Suffix}" >> $env:GITHUB_OUTPUT
+          echo "llvm_config=${{ github.workspace }}/BinaryCache/0/bin/llvm-config${Suffix}" >> $env:GITHUB_OUTPUT
+          echo "clang_pseudo_gen=${{ github.workspace }}/BinaryCache/0/bin/clang-pseudo-gen${Suffix}" >> $env:GITHUB_OUTPUT
+          echo "clang_tidy_confusable_chars_gen=${{ github.workspace }}/BinaryCache/0/bin/clang-tidy-confusable-chars-gen${Suffix}" >> $env:GITHUB_OUTPUT
+          echo "swift_def_to_strings_converter=${{ github.workspace }}/BinaryCache/0/bin/swift-def-to-strings-converter${Suffix}" >> $env:GITHUB_OUTPUT
+          echo "swift_serialize_diagnostics=${{ github.workspace }}/BinaryCache/0/bin/swift-serialize-diagnostics${Suffix}" >> $env:GITHUB_OUTPUT
+          echo "swift_compatibility_symbols=${{ github.workspace }}/BinaryCache/0/bin/swift-compatibility-symbols${Suffix}" >> $env:GITHUB_OUTPUT
```
compnerd, line 506: I wonder if we can get away with a heredoc instead of the multiple `echo`s. However, we can do that as a follow-up.

compnerd (approved):

PR #824, opened 2024-09-04
[Mac build] Add zlib to the Mac build

Steelskin:
* Make zlib version an input so it only needs to be set once.
* Add matrix for Windows and Darwin targets. For now, Darwin does not build Android targets.


compnerd (approved):
Is there a way to reduce the split and have a single job for zlib, curl, libxml2 that just extends the build matrix?

Steelskin:
> Is there a way to reduce the split and have a single job for zlib, curl, libxml2 that just extends the build matrix?

We can definitely merge all 3 jobs into one since they pretty much do the same thing except with different repos. It would add a bit more build time since they would no longer run in parallel but they are pretty fast already and the other jobs are already waiting on the `compilers` job already.
I am a bit reluctant to do these kind of larger changes at the same time because of the added risks but I'm making a note to improve this once the Mac build is complete.

PR #829, opened 2024-09-06
Merge zlib, curl and libxml2 into a single job

### LINKED ISSUE #825: Merge zlib, curl and libxml2 into a single job
Right now, `zlib`, `curl` and `libxml2` are built as separate parallel jobs. However, the jobs depending on the build artifacts from these are already waiting on the `compilers` job to complete, making the parallelism pointless. These 3 jobs should be merged and run as a single job, tentatively called `target-libs`.

Steelskin:
These 3 jobs are all configured the same way and do not need to be separate jobs each spinning a new build bot instance.

Fixes #825

### DIFF
--- a/.github/workflows/swift-toolchain.yml
+++ b/.github/workflows/swift-toolchain.yml
@@ -1004,7 +1004,7 @@ jobs:
         uses: hendrikmuhs/ccache-action@2e0e89e8d74340a03f75d58d02aae4c5ee1b15c6
         with:
           max-size: 100M
-          key: ${{ steps.workspace_hash.outputs.hash }}-${{ matrix.os }}-${{ matrix.arch }}-target-libs
+          key: ${{ steps.workspace_hash.outputs.hash }}-${{ matrix.os }}-${{ matrix.arch }}-dependencies
           variant: sccache
 
       - uses: nttld/setup-ndk@v1
--- a/.github/workflows/swift-toolchain.yml
+++ b/.github/workflows/swift-toolchain.yml
@@ -969,23 +969,17 @@ jobs:
           symbolsFolder: ${{ github.workspace }}/BinaryCache/1
           searchPattern: '**/*.exe'
 
-  zlib:
+  target_libs:
     runs-on: ${{ inputs.default_build_runner }}
 
     strategy:
       fail-fast: false
       matrix: ${{ fromJSON(inputs.target_matrix) }}
 
-    name: ${{ matrix.os }} ${{ matrix.arch }} zlib
+    name: ${{ matrix.os }} ${{ matrix.arch }} Target libs
 
     steps:
-      - uses: actions/checkout@v4
-        with:
-          repository: madler/zlib
-          ref: ${{ inputs.zlib_revision }}
-          path: ${{ github.workspace }}/SourceCache/zlib
-          show-progress: false
-
+      # Setup build tools.
       - uses: compnerd/gha-setup-vsdevenv@main
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/swift-toolchain.yml
```diff
@@ -969,23 +969,17 @@ jobs:
           symbolsFolder: ${{ github.workspace }}/BinaryCache/1
           searchPattern: '**/*.exe'
 
-  zlib:
+  target_libs:
```
compnerd, line 972:
```suggestion
  dependencies:
```

.github/workflows/swift-toolchain.yml
```diff
@@ -969,23 +969,17 @@ jobs:
...
     runs-on: ${{ inputs.default_build_runner }}
 
     strategy:
       fail-fast: false
       matrix: ${{ fromJSON(inputs.target_matrix) }}
 
-    name: ${{ matrix.os }} ${{ matrix.arch }} zlib
+    name: ${{ matrix.os }} ${{ matrix.arch }} Target libs
```
compnerd, line 979:
```suggestion
    name: ${{ matrix.os }} ${{ matrix.arch }} Dependencies
```

.github/workflows/swift-toolchain.yml
```diff
@@ -1010,14 +1004,22 @@ jobs:
         uses: hendrikmuhs/ccache-action@2e0e89e8d74340a03f75d58d02aae4c5ee1b15c6
         with:
           max-size: 100M
-          key: ${{ steps.workspace_hash.outputs.hash }}-${{ matrix.os }}-${{ matrix.arch }}-zlib
+          key: ${{ steps.workspace_hash.outputs.hash }}-${{ matrix.os }}-${{ matrix.arch }}-target-libs
```
compnerd, line 1007:
```suggestion
          key: ${{ steps.workspace_hash.outputs.hash }}-${{ matrix.os }}-${{ matrix.arch }}-dependencies
```

.github/workflows/swift-toolchain.yml
```diff
@@ -1755,7 +1668,7 @@ jobs:
   sdk:
     # TODO: Build this on macOS or make an equivalent Mac-only job
     if: inputs.build_os == 'Windows'
-    needs: [libxml2, curl, zlib, compilers, cmark_gfm, stdlib, macros]
+    needs: [target_libs, compilers, cmark_gfm, stdlib, macros]
```
compnerd, line 1671:
```suggestion
    needs: [dependencies, compilers, cmark_gfm, stdlib, macros]
```

compnerd:
I'm not sure if we want to necessarily merge these into a single job, there is some value in the parallelism still. I was thinking more that we want to merge the Darwin and Windows builds into a single instance rather than a separate set for Darwin and a separate set for Windows.

Steelskin:
> I'm not sure if we want to necessarily merge these into a single job, there is some value in the parallelism still. I was thinking more that we want to merge the Darwin and Windows builds into a single instance rather than a separate set for Darwin and a separate set for Windows.

Ah you would like to re-merge the flows instead? That's something that should be doable after the mac work is finished. I think this is still a good change because this is only consumed by the sdk job, which is happening after the compilers job so we don't get much from parallelization and this should be a bit cheaper to run since we don't need to spawn new instances for every sub-job.

compnerd:
I am unsure about the merge because I'm thinking that we may want to consume zlib and libxml2 in the compilers build as well, but not CURL.

PR #830, opened 2024-09-19
[ds2] update ds2 revision to nightly-2024-09-19

andrurogerz:
Updates the swift toolchain to build a newer version of ds2 that contains a number of fixes.


andrurogerz:
revised to the 9-19 build because that's the one that *actually* has the fixes in it

PR #832, opened 2024-09-27
Set cflags matrix variable properly

Steelskin:
The variable had been erroneously names `ccflags`.


compnerd (approved):
Do the usage need to be updated as well?

Steelskin:
> Do the usage need to be updated as well?

No, the error was in the matrix definition, not the usage.

PR #841, opened 2024-11-04
Fixes to get main working downstream again

hyp:


hyp:
1:

    Do not attempt NuGet registration
    
    This is now failing as NuGet is registered:
    
    ```
    Register-PackageSource: D:\r\_work\_temp\c04a7ca5-aa41-491b-b9ad-0e09fb799d0c.ps1:4
    Line |
       4 |      Register-PackageSource -Name NuGet -Location https://www.nuget.or …
         |      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
         | Package Source 'NuGet' exists.
    ```
    
    https://github.com/thebrowsercompany/swift-build/actions/runs/11656344808/job/32454421580

2:

    GHA workflow: add swift testing package name
    
    The build of swift testing now requires a specified package name,
    as swift testing uses the 'package' keyword. swift_testing the
    name used in the cmake build as well.

3:

    Update base swift compiler to 20231016.5
    
    The 20231016.5 is a new build based on 20231016.0
    that has the fix for `static_assert(false` failure
    in Clang that was failing to compile Swift with updated
    MSVC c++ headers.

compnerd (approved):
I'm fine with this. I think that this might break the builds on custom runners (I think that was why @kendalharland added the nuget source). But let's get the GHA build working first, and work towards making the custom builders identical to the GHA agents.

PR #843, opened 2024-11-07
[Mac Build] Add stdlib build

Steelskin:
* Adapt stdlib job to build for both Windows and Mac.

### DIFF
--- a/.github/workflows/build-toolchain.yml
+++ b/.github/workflows/build-toolchain.yml
@@ -379,7 +379,6 @@ jobs:
                   "cflags": "${{ steps.context.outputs.WINDOWS_CMAKE_C_FLAGS }}",
                   "cxx": "cl",
                   "cxxflags": "${{ steps.context.outputs.WINDOWS_CMAKE_CXX_FLAGS }}",
-                  "swiftflags": "${{ steps.context.outputs.WINDOWS_CMAKE_Swift_FLAGS }}",
                   "cmake_linker_flags": "-D CMAKE_EXE_LINKER_FLAGS=\"${{ steps.context.outputs.WINDOWS_CMAKE_EXE_LINKER_FLAGS }}\" -D CMAKE_SHARED_LINKER_FLAGS=\"${{ steps.context.outputs.WINDOWS_CMAKE_SHARED_LINKER_FLAGS }}\" -D CMAKE_STATIC_LIBRARY_PREFIX_Swift=lib",
                   "extra_flags": "-D CMAKE_SYSTEM_PROCESSOR=AMD64"
                 },
@@ -393,7 +392,6 @@ jobs:
                   "cflags": "${{ steps.context.outputs.WINDOWS_CMAKE_C_FLAGS }}",
                   "cxx": "cl",
                   "cxxflags": "${{ steps.context.outputs.WINDOWS_CMAKE_CXX_FLAGS }}",
-                  "swiftflags": "${{ steps.context.outputs.WINDOWS_CMAKE_Swift_FLAGS }}",
                   "cmake_linker_flags": "-D CMAKE_EXE_LINKER_FLAGS=\"${{ steps.context.outputs.WINDOWS_CMAKE_EXE_LINKER_FLAGS }}\" -D CMAKE_SHARED_LINKER_FLAGS=\"${{ steps.context.outputs.WINDOWS_CMAKE_SHARED_LINKER_FLAGS }}\" -D CMAKE_STATIC_LIBRARY_PREFIX_Swift=lib",
                   "extra_flags": "-D CMAKE_SYSTEM_PROCESSOR=ARM64"
                 }
@@ -412,7 +410,6 @@ jobs:
                   "cflags": "${{ steps.context.outputs.WINDOWS_CMAKE_C_FLAGS }}",
                   "cxx": "cl",
                   "cxxflags": "${{ steps.context.outputs.WINDOWS_CMAKE_CXX_FLAGS }}",
-                  "swiftflags": "${{ steps.context.outputs.WINDOWS_CMAKE_Swift_FLAGS }}",
                   "cmake_linker_flags": "-D CMAKE_EXE_LINKER_FLAGS=\"${{ steps.context.outputs.WINDOWS_CMAKE_EXE_LINKER_FLAGS }}\" -D CMAKE_SHARED_LINKER_FLAGS=\"${{ steps.context.outputs.WINDOWS_CMAKE_SHARED_LINKER_FLAGS }}\" -D CMAKE_STATIC_LIBRARY_PREFIX_Swift=lib",
                   "extra_flags": "-D CMAKE_SYSTEM_PROCESSOR=ARM64"
                 }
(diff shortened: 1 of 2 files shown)


.github/workflows/swift-toolchain.yml
```diff
@@ -1573,121 +1573,21 @@ jobs:
           path: ${{ github.workspace }}/BuildRoot/Library/libxml2-${{ inputs.libxml2_version }}/usr
 
   stdlib:
-    # TODO: Build this on macOS or make an equivalent Mac-only job
-    if: inputs.build_os == 'Windows'
-    needs: [compilers, cmark_gfm]
+    needs: [compilers]
```
compnerd, line 1576: Compilers will need the CMark GFM to execute, why is the dependency removed?

.github/workflows/swift-toolchain.yml
```diff
@@ -1712,14 +1612,25 @@ jobs:
           ref: ${{ inputs.swift_experimental_string_processing_revision }}
           path: ${{ github.workspace }}/SourceCache/swift-experimental-string-processing
           show-progress: false
+      - uses: actions/checkout@v4
+        if: matrix.os == 'Darwin'
+        with:
+          path: ${{ github.workspace }}/SourceCache/swift-build
```
compnerd, line 1618: Can you use the explicit revision to checkout at?

.github/workflows/swift-toolchain.yml
```diff
@@ -1728,84 +1639,113 @@ jobs:
...
+
+          # Export the configuration.
+          Write-Output "swiftc=${SwiftC}" | Out-File -FilePath ${env:GITHUB_OUTPUT} -Encoding utf8 -Append
+          Write-Output "swiftflags=${SwiftFlags}" | Out-File -FilePath ${env:GITHUB_OUTPUT} -Encoding utf8 -Append
+          Write-Output "install-dir=${InstallDir}" | Out-File -FilePath ${env:GITHUB_OUTPUT} -Encoding utf8 -Append
+          Write-Output "ndk-flags=${NdkFlags}" | Out-File -FilePath ${env:GITHUB_OUTPUT} -Encoding utf8 -Append
+          Write-Output "stdlib-cc=${StdlibCc}" | Out-File -FilePath ${env:GITHUB_OUTPUT} -Encoding utf8 -Append
+          Write-Output "stdlib-cxx=${StdlibCxx}" | Out-File -FilePath ${env:GITHUB_OUTPUT} -Encoding utf8 -Append
+          Write-Output "stdlib-flags=${StdlibFlags}" | Out-File -FilePath ${env:GITHUB_OUTPUT} -Encoding utf8 -Append
+          Write-Output "stdlib-cache-file=${StdlibCacheFile}" | Out-File -FilePath ${env:GITHUB_OUTPUT} -Encoding utf8 -Append
```
compnerd, line 1703: Can you use a heredoc instead?

.github/workflows/swift-toolchain.yml
```diff
@@ -1728,84 +1639,113 @@ jobs:
...
+          } else {
+            $SwiftC = "${{ github.workspace }}/BinaryCache/Library/Developer/Toolchains/unknown-Asserts-development.xctoolchain/usr/bin/swiftc"
+            $InstallDir = "${{ github.workspace }}/BuildRoot/Library/Developer/Toolchains/unknown-Asserts-development.xctoolchain/usr"
+            $NdkFlags = ""
+
+            $StdlibCc = "${{ github.workspace }}/BinaryCache/Library/Developer/Toolchains/unknown-Asserts-development.xctoolchain/usr/bin/clang"
+            $StdlibCxx = "${{ github.workspace }}/BinaryCache/Library/Developer/Toolchains/unknown-Asserts-development.xctoolchain/usr/bin/clang++"
+            $StdlibFlags = ""
+            $StdlibCacheFile = "${{ github.workspace }}/SourceCache/swift-build/Runtime-${{ matrix.platform }}-${{ matrix.arch }}.cmake"
+          }
```
compnerd, line 1693: Can we push this down into the workflow through the matrix and simplify this somehow?

compnerd (commented):
Any chance we could re-use or try to upstream the cmake caches?

.github/workflows/swift-toolchain.yml
Steelskin: I moved these to the matrices. It's a bit simpler now.

.github/workflows/swift-toolchain.yml
Steelskin: This has been removed.

.github/workflows/build-toolchain.yml
```diff
@@ -455,73 +455,164 @@ jobs:
...
                   "cflags": "${{ steps.context.outputs.WINDOWS_CMAKE_C_FLAGS }}",
                   "cxx": "cl",
                   "cxxflags": "${{ steps.context.outputs.WINDOWS_CMAKE_CXX_FLAGS }}",
                   "swiftflags": "${{ steps.context.outputs.WINDOWS_CMAKE_Swift_FLAGS }}",
-                  "extra_flags": "-D CMAKE_SYSTEM_PROCESSOR=AMD64"
+                  "cmake_linker_flags": "-D CMAKE_EXE_LINKER_FLAGS=\"${{ steps.context.outputs.WINDOWS_CMAKE_EXE_LINKER_FLAGS }}\" -D CMAKE_SHARED_LINKER_FLAGS=\"${{ steps.context.outputs.WINDOWS_CMAKE_EXE_LINKER_FLAGS }}\"",
+                  "cmake_ndk_flags": "",
+                  "extra_flags": "-D CMAKE_SYSTEM_PROCESSOR=AMD64",
+                  "build_cc": "${HostToolchain}/usr/bin/clang-cl.exe",
+                  "build_cxx": "${HostToolchain}/usr/bin/clang-cl.exe",
```
compnerd, line 472: These are confusing. The build C/C++ compilers change between phases.

Steelskin: That was already the case prior to these changes. This is making the compiler used (system vs. built) more explicit.
Should we use the same compiler everywhere? I don't know if the long-term goal is to use `clang-cl` or MSVC `cl`.

compnerd: > That was already the case prior to these changes

Sure, the compiler changing between phases was already the case, but we didn't label clang as the build compiler and cl as something else. This shouldn't be called `build_cc`, unless you want to define this path 25-75 times, once for each of the repository/target case.

.github/workflows/swift-toolchain.yml
compnerd: It should be though - we need to download and extract it. If we don't the compiler will not run.

.github/workflows/swift-toolchain.yml
```diff
@@ -1732,75 +1642,67 @@ jobs:
...
           }
+
           cmake -B ${{ github.workspace }}/BinaryCache/llvm `
                 -D CMAKE_BUILD_TYPE=Release `
                 -D CMAKE_C_COMPILER=${{ matrix.cc }} `
                 -D CMAKE_C_FLAGS="${{ matrix.cflags }}" `
                 -D CMAKE_CXX_COMPILER=${{ matrix.cxx }} `
                 -D CMAKE_CXX_FLAGS="${{ matrix.cxxflags }}" `
-                -D CMAKE_INSTALL_PREFIX=${{ github.workspace }}/BuildRoot/Library/Developer/Platforms/${{ matrix.os }}.platform/Developer/SDKs/${{ matrix.os }}.sdk/usr `
+                -D CMAKE_INSTALL_PREFIX=${InstallDir} `
```
compnerd, line 1662: Why does this need to be `cygpath`'ed suddenly?

.github/workflows/build-toolchain.yml
Steelskin: I have removed them and am now setting the compiler manually before running configure.

.github/workflows/swift-toolchain.yml
Steelskin: A bad update on my part. Anyway, this is now entirely set up from the matrix.

compnerd (approved):

PR #847, opened 2024-11-19
Add the native ARM64 toolchain build to the main CI

hjyamauchi:
- Introduce the concept of ARM64 CI runners. Use the cirun ARM64 machines as the ARM64 runners.
- Split the windows_build job into X64 and ARM64.
- Add the Win/ARM64 build on the Win/ARM64 runner (but no Win/X64 build on Win/ARM64 runner) while retaining the existing Win/X64 and Win/ARM64 builds on the Win/X64 runner.
- Split some of strategy matrices between X64 and ARM64.
- Append the "... (building on <os> <arch>)" to the job names to clarify what's building on what runner.
- Do not release the natively-released Win/ARM64 installer.exe while retaining the existing cross-compiled Win/ARM64 and Win/X64 installer.exe as the published release.
- Exclude the Android-related components including ds2 in the native Win/ARM64 build due to the non-existence of the native Windows/ARM64 Android NDK.
- Other misc changes to get the jobs to be able to run natively on Win/ARM64 (e.g., overriding the build/host arch to arm64 for the vsdevenv or the swift toolchain setup actions).
- The native Win/ARM64 build should currently fail in the compilers job due to the static_assert issue in the pinned toolchain but the whole main CI run should succeed and produce builds as the `release` job depends only on `windows-build-x64`.

### DIFF
--- a/.github/workflows/build-toolchain.yml
+++ b/.github/workflows/build-toolchain.yml
@@ -36,6 +36,17 @@ on:
         default: true
         required: false
 
+      create_snapshot:
+        description: 'Create Snapshot'
+        type: boolean
+        default: true
+        required: false
+
+      build_android:
+        required: false
+        default: false
+        type: boolean
+
   workflow_call:
     inputs:
       swift_version:
@@ -71,15 +82,41 @@ on:
         default: true
         required: false
 
-      windows_default_runner:
-        description: 'Build runner'
+      create_snapshot:
+        description: 'Create Snapshot'
+        type: boolean
+        default: true
+        required: false
+
+      windows_x64_default_runner:
+        description: 'X64 Build runner'
         required: false
         type: string
 
-      windows_compilers_runner:
-        description: 'Build runner for `compilers` job'
-        type: string
+      windows_arm64_default_runner:
+        description: 'ARM64 Build runner'
         required: false
+        type: string
+
+      windows_x64_compilers_runner:
+        description: 'X64 Build runner for `compilers` job'
+        required: false
+        type: string
+
+      windows_arm64_compilers_runner:
+        description: 'ARM64 Build runner for `compilers` job'
+        required: false
+        type: string
+
+      windows_build_arch:
+        description: 'Windows build architecture'
+        required: false
+        type: string
+
+      build_android:
+        required: false
+        default: false
+        type: boolean
 
     secrets:
       SYMBOL_SERVER_PAT:
@@ -153,12 +190,19 @@ jobs:
       signed: ${{ steps.context.outputs.signed }}
       swift_version: ${{ steps.context.outputs.swift_version }}
       swift_tag: ${{ steps.context.outputs.swift_tag }}
-      windows_build_runner: ${{ steps.context.outputs.windows_build_runner }}
-      windows_compilers_runner: ${{ steps.context.outputs.windows_compilers_runner }}
+      windows_arm64_build_runner: ${{ steps.context.outputs.windows_arm64_build_runner }}
+      windows_arm64_compilers_runner: ${{ steps.context.outputs.windows_arm64_compilers_runner }}
+      windows_build_arch: ${{ steps.context.outputs.windows_build_arch }}
+      windows_build_cpu: ${{ steps.context.outputs.windows_build_cpu }}
+      windows_x64_build_runner: ${{ steps.context.outputs.windows_x64_build_runner }}
+      windows_x64_compilers_runner: ${{ steps.context.outputs.windows_x64_compilers_runner }}
       mac_build_runner: ${{ steps.context.outputs.mac_build_runner }}
-      windows_host_matrix: ${{ steps.setup-matrix.outputs.windows_host_matrix }}
-      windows_build_matrix: ${{ steps.setup-matrix.outputs.windows_build_matrix }}
-      windows_target_matrix: ${{ steps.setup-matrix.outputs.windows_target_matrix }}
+      windows_arm64_build_matrix: ${{ steps.setup-matrix.outputs.windows_arm64_build_matrix }}
+      windows_arm64_host_matrix: ${{ steps.setup-matrix.outputs.windows_arm64_host_matrix }}
+      windows_arm64_target_matrix: ${{ steps.setup-matrix.outputs.windows_arm64_target_matrix }}
+      windows_x64_build_matrix: ${{ steps.setup-matrix.outputs.windows_x64_build_matrix }}
+      windows_x64_host_matrix: ${{ steps.setup-matrix.outputs.windows_x64_host_matrix }}
+      windows_x64_target_matrix: ${{ steps.setup-matrix.outputs.windows_x64_target_matrix }}
       darwin_host_matrix: ${{ steps.setup-matrix.outputs.darwin_host_matrix }}
       darwin_build_matrix: ${{ steps.setup-matrix.outputs.darwin_build_matrix }}
       darwin_target_matrix: ${{ steps.setup-matrix.outputs.darwin_target_matrix }}
@@ -284,8 +328,18 @@ jobs:
             fi
           fi
 
-          echo windows_build_runner=${{ inputs.windows_default_runner || vars.WINDOWS_BUILD_RUNNER || 'windows-latest' }} >> ${GITHUB_OUTPUT}
-          echo windows_compilers_runner=${{ inputs.windows_compilers_runner || inputs.windows_default_runner || vars.COMPILERS_BUILD_RUNNER || vars.WINDOWS_BUILD_RUNNER || 'windows-latest' }} >> ${GITHUB_OUTPUT}
+          echo windows_x64_build_runner=${{ inputs.windows_x64_default_runner || vars.WINDOWS_BUILD_RUNNER || 'windows-latest' }} >> ${GITHUB_OUTPUT}
+          echo windows_x64_compilers_runner=${{ inputs.windows_x64_compilers_runner || inputs.windows_x64_default_runner || vars.WINDOWS_X64_COMPILERS_BUILD_RUNNER || vars.WINDOWS_X64_DEFAULT_BUILD_RUNNER || 'windows-latest' }} >> ${GITHUB_OUTPUT}
+          echo windows_arm64_build_runner=${{ inputs.windows_arm64_default_runner || vars.WINDOWS_ARM64_DEFAULT_BUILD_RUNNER || 'windows-latest' }} >> ${GITHUB_OUTPUT}
+          echo windows_arm64_compilers_runner=${{ inputs.windows_arm64_compilers_runner || inputs.windows_arm64_default_runner || vars.WINDOWS_ARM64_COMPILERS_BUILD_RUNNER || vars.WINDOWS_ARM64_DEFAULT_BUILD_RUNNER || 'windows-latest' }} >> ${GITHUB_OUTPUT}
+          if [[ "${{ inputs.windows_build_arch }}" == "arm64" ]] ; then
+            echo windows_build_arch=arm64 >> ${GITHUB_OUTPUT}
+            echo windows_build_cpu=arm64 >> ${GITHUB_OUTPUT}
+          else
+            echo windows_build_arch=amd64 >> ${GITHUB_OUTPUT}
+            echo windows_build_cpu=x64 >> ${GITHUB_OUTPUT}
+          fi
+
           # TODO: Make the mac runner configurable.
           echo mac_build_runner=macos-latest >> ${GITHUB_OUTPUT}
 
@@ -293,6 +347,7 @@ jobs:
           echo ANDROID_NDK_VERSION=r26b >> ${GITHUB_OUTPUT}
 
       - uses: actions/upload-artifact@v4
+        if: inputs.create_snapshot == true
         with:
           name: stable.xml
           path: stable.xml
@@ -301,7 +356,56 @@ jobs:
       - name: Setup matrix
         id: setup-matrix
         env:
-          WINDOWS_HOST_MATRIX: >-
+          WINDOWS_X64_HOST_MATRIX: >-
+            {
+              "include": [
+                {
+                  "arch": "amd64",
+                  "cpu": "x86_64",
+                  "triple": "x86_64-unknown-windows-msvc",
+                  "compiler_target": "x86_64-unknown-windows-msvc",
+                  "os": "Windows",
+                  "cc": "cl",
+                  "cflags": "${{ steps.context.outputs.WINDOWS_CMAKE_C_FLAGS }}",
+                  "cxx": "cl",
+                  "cxxflags": "${{ steps.context.outputs.WINDOWS_CMAKE_CXX_FLAGS }}",
+                  "swiftflags": "${{ steps.context.outputs.WINDOWS_CMAKE_Swift_FLAGS }}",
+                  "extra_flags": "-D CMAKE_SYSTEM_PROCESSOR=AMD64 -D CMAKE_MT=mt"
+                },
+                {
+                  "arch": "arm64",
+                  "cpu": "aarch64",
+                  "triple": "aarch64-unknown-windows-msvc",
+                  "compiler_target": "aarch64-unknown-windows-msvc",
+                  "os": "Windows",
+                  "cc": "cl",
+                  "cflags": "${{ steps.context.outputs.WINDOWS_CMAKE_C_FLAGS }}",
+                  "cxx": "cl",
+                  "cxxflags": "${{ steps.context.outputs.WINDOWS_CMAKE_CXX_FLAGS }}",
+                  "swiftflags": "${{ steps.context.outputs.WINDOWS_CMAKE_Swift_FLAGS }}",
+                  "extra_flags": "-D CMAKE_SYSTEM_PROCESSOR=ARM64 -D CMAKE_MT=mt"
+                }
+              ]
+            }
+          WINDOWS_ARM64_HOST_MATRIX: >-
+            {
+              "include": [
+                {
+                  "arch": "arm64",
+                  "cpu": "aarch64",
+                  "triple": "aarch64-unknown-windows-msvc",
+                  "compiler_target": "aarch64-unknown-windows-msvc",
+                  "os": "Windows",
+                  "cc": "cl",
+                  "cflags": "${{ steps.context.outputs.WINDOWS_CMAKE_C_FLAGS }}",
+                  "cxx": "cl",
+                  "cxxflags": "${{ steps.context.outputs.WINDOWS_CMAKE_CXX_FLAGS }}",
+                  "swiftflags": "${{ steps.context.outputs.WINDOWS_CMAKE_Swift_FLAGS }}",
+                  "extra_flags": "-D CMAKE_SYSTEM_PROCESSOR=ARM64 -D CMAKE_MT=mt"
+                }
+              ]
+            }
+          WINDOWS_X64_BUILD_MATRIX: >-
             {
               "include": [
                 {
@@ -314,7 +418,12 @@ jobs:
                   "cxxflags": "${{ steps.context.outputs.WINDOWS_CMAKE_CXX_FLAGS }}",
                   "swiftflags": "${{ steps.context.outputs.WINDOWS_CMAKE_Swift_FLAGS }}",
                   "extra_flags": "-D CMAKE_SYSTEM_PROCESSOR=AMD64 -D CMAKE_MT=mt"
-                },
+                }
+              ]
+            }
+          WINDOWS_ARM64_BUILD_MATRIX: >-
+            {
+              "include": [
                 {
                   "arch": "arm64",
                   "compiler_target": "aarch64-unknown-windows-msvc",
@@ -328,22 +437,7 @@ jobs:
                 }
               ]
             }
-          WINDOWS_BUILD_MATRIX: >-
-            {
-              "include": [
-                {
-                  "arch": "amd64",
-                  "os": "Windows",
-                  "cc": "cl",
-                  "cflags": "${{ steps.context.outputs.WINDOWS_CMAKE_C_FLAGS }}",
-                  "cxx": "cl",
-                  "cxxflags": "${{ steps.context.outputs.WINDOWS_CMAKE_CXX_FLAGS }}",
-                  "swiftflags": "${{ steps.context.outputs.WINDOWS_CMAKE_Swift_FLAGS }}",
-                  "extra_flags": "-D CMAKE_SYSTEM_PROCESSOR=AMD64 -D CMAKE_MT=mt"
-                }
-              ]
-            }
-          WINDOWS_TARGET_MATRIX: >-
+          WINDOWS_X64_TARGET_MATRIX: >-
             {
               "include": [
                 {
@@ -418,6 +512,41 @@ jobs:
                 }
               ]
             }
+          WINDOWS_ARM64_TARGET_MATRIX: >-
+            {
+              "include": [
+                {
+                  "arch": "amd64",
+                  "os": "Windows",
+                  "cc": "cl",
+                  "cflags": "${{ steps.context.outputs.WINDOWS_CMAKE_C_FLAGS }}",
+                  "cxx": "cl",
+                  "cxxflags": "${{ steps.context.outputs.WINDOWS_CMAKE_CXX_FLAGS }}",
+                  "swiftflags": "${{ steps.context.outputs.WINDOWS_CMAKE_Swift_FLAGS }}",
+                  "extra_flags": "-D CMAKE_SYSTEM_PROCESSOR=AMD64 -D CMAKE_MT=mt"
+                },
+                {
+                  "arch": "arm64",
+                  "os": "Windows",
+                  "cc": "cl",
+                  "cflags": "${{ steps.context.outputs.WINDOWS_CMAKE_C_FLAGS }}",
+                  "cxx": "cl",
+                  "cxxflags": "${{ steps.context.outputs.WINDOWS_CMAKE_CXX_FLAGS }}",
+                  "swiftflags": "${{ steps.context.outputs.WINDOWS_CMAKE_Swift_FLAGS }}",
+                  "extra_flags": "-D CMAKE_SYSTEM_PROCESSOR=ARM64 -D CMAKE_MT=mt"
+                },
+                {
+                  "arch": "x86",
+                  "os": "Windows",
+                  "cc": "cl",
+                  "cflags": "${{ steps.context.outputs.WINDOWS_CMAKE_C_FLAGS }}",
+                  "cxx": "cl",
+                  "cxxflags": "${{ steps.context.outputs.WINDOWS_CMAKE_CXX_FLAGS }}",
+                  "swiftflags": "",
+                  "extra_flags": "-D CMAKE_SYSTEM_PROCESSOR=X86 -D CMAKE_MT=mt"
+                }
+              ]
+            }
           DARWIN_HOST_MATRIX: >-
             {
               "include": [
@@ -486,9 +615,12 @@ jobs:
               ]
             }
         run: |
-          echo "windows_host_matrix=$(jq -r -c '.' <<< ${WINDOWS_HOST_MATRIX})" >> ${GITHUB_OUTPUT}
-          echo "windows_build_matrix=$(jq -r -c '.' <<< ${WINDOWS_BUILD_MATRIX})" >> ${GITHUB_OUTPUT}
-          echo "windows_target_matrix=$(jq -r -c '.' <<< ${WINDOWS_TARGET_MATRIX})" >> ${GITHUB_OUTPUT}
+          echo "windows_x64_host_matrix=$(jq -r -c '.' <<< ${WINDOWS_X64_HOST_MATRIX})" >> ${GITHUB_OUTPUT}
+          echo "windows_arm64_host_matrix=$(jq -r -c '.' <<< ${WINDOWS_ARM64_HOST_MATRIX})" >> ${GITHUB_OUTPUT}
+          echo "windows_x64_build_matrix=$(jq -r -c '.' <<< ${WINDOWS_X64_BUILD_MATRIX})" >> ${GITHUB_OUTPUT}
+          echo "windows_arm64_build_matrix=$(jq -r -c '.' <<< ${WINDOWS_ARM64_BUILD_MATRIX})" >> ${GITHUB_OUTPUT}
+          echo "windows_x64_target_matrix=$(jq -r -c '.' <<< ${WINDOWS_X64_TARGET_MATRIX})" >> ${GITHUB_OUTPUT}
+          echo "windows_arm64_target_matrix=$(jq -r -c '.' <<< ${WINDOWS_ARM64_TARGET_MATRIX})" >> ${GITHUB_OUTPUT}
           echo "darwin_host_matrix=$(jq -r -c '.' <<< ${DARWIN_HOST_MATRIX})" >> ${GITHUB_OUTPUT}
           echo "darwin_build_matrix=$(jq -r -c '.' <<< ${DARWIN_BUILD_MATRIX})" >> ${GITHUB_OUTPUT}
           echo "darwin_target_matrix=$(jq -r -c '.' <<< ${DARWIN_TARGET_MATRIX})" >> ${GITHUB_OUTPUT}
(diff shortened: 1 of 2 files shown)


.github/workflows/build-toolchain.yml
```diff
@@ -153,12 +163,17 @@ jobs:
...
       mac_build_runner: ${{ steps.context.outputs.mac_build_runner }}
-      windows_host_matrix: ${{ steps.setup-matrix.outputs.windows_host_matrix }}
-      windows_build_matrix: ${{ steps.setup-matrix.outputs.windows_build_matrix }}
-      windows_target_matrix: ${{ steps.setup-matrix.outputs.windows_target_matrix }}
+      windows_x64_host_matrix: ${{ steps.setup-matrix.outputs.windows_x64_host_matrix }}
+      windows_arm64_host_matrix: ${{ steps.setup-matrix.outputs.windows_arm64_host_matrix }}
+      windows_x64_build_matrix: ${{ steps.setup-matrix.outputs.windows_x64_build_matrix }}
+      windows_arm64_build_matrix: ${{ steps.setup-matrix.outputs.windows_arm64_build_matrix }}
+      windows_x64_target_matrix: ${{ steps.setup-matrix.outputs.windows_x64_target_matrix }}
+      windows_arm64_target_matrix: ${{ steps.setup-matrix.outputs.windows_arm64_target_matrix }}
```
compnerd, line 176: Please sort these lexicographically.

.github/workflows/build-toolchain.yml
```diff
@@ -284,8 +299,10 @@ jobs:
             fi
           fi
 
-          echo windows_build_runner=${{ inputs.windows_default_runner || vars.WINDOWS_BUILD_RUNNER || 'windows-latest' }} >> ${GITHUB_OUTPUT}
-          echo windows_compilers_runner=${{ inputs.windows_compilers_runner || inputs.windows_default_runner || vars.COMPILERS_BUILD_RUNNER || vars.WINDOWS_BUILD_RUNNER || 'windows-latest' }} >> ${GITHUB_OUTPUT}
+          echo windows_x64_build_runner=${{ inputs.windows_x64_default_runner || vars.WINDOWS_BUILD_RUNNER || 'windows-latest' }} >> ${GITHUB_OUTPUT}
+          echo windows_x64_compilers_runner=${{ inputs.windows_x64_compilers_runner || inputs.windows_x64_default_runner || vars.COMPILERS_BUILD_RUNNER || vars.WINDOWS_BUILD_RUNNER || 'windows-latest' }} >> ${GITHUB_OUTPUT}
+          echo windows_arm64_build_runner=${{ inputs.windows_arm64_default_runner }} >> ${GITHUB_OUTPUT}
+          echo windows_arm64_compilers_runner=${{ inputs.windows_arm64_compilers_runner || inputs.windows_arm64_default_runner }} >> ${GITHUB_OUTPUT}
```
compnerd, line 305: We should be able to control these similar to the x64 builds. Currently, this will break anyone building with a fork of this repository.

.github/workflows/build-toolchain.yml
```diff
@@ -486,23 +577,26 @@ jobs:
...
+          echo "windows_x64_build_matrix=$(jq -r -c '.' <<< ${WINDOWS_X64_BUILD_MATRIX})" >> ${GITHUB_OUTPUT}
+          echo "windows_arm64_build_matrix=$(jq -r -c '.' <<< ${WINDOWS_ARM64_BUILD_MATRIX})" >> ${GITHUB_OUTPUT}
+          echo "windows_x64_target_matrix=$(jq -r -c '.' <<< ${WINDOWS_X64_TARGET_MATRIX})" >> ${GITHUB_OUTPUT}
+          echo "windows_arm64_target_matrix=$(jq -r -c '.' <<< ${WINDOWS_ARM64_TARGET_MATRIX})" >> ${GITHUB_OUTPUT}
           echo "darwin_host_matrix=$(jq -r -c '.' <<< ${DARWIN_HOST_MATRIX})" >> ${GITHUB_OUTPUT}
           echo "darwin_build_matrix=$(jq -r -c '.' <<< ${DARWIN_BUILD_MATRIX})" >> ${GITHUB_OUTPUT}
           echo "darwin_target_matrix=$(jq -r -c '.' <<< ${DARWIN_TARGET_MATRIX})" >> ${GITHUB_OUTPUT}
 
-  windows-build:
+  windows-build-x64:
```
compnerd, line 590: Can we use a matrix strategy here instead?

.github/workflows/build-toolchain.yml
```diff
@@ -662,17 +827,17 @@ jobs:
...
       - uses: actions/download-artifact@v4
         with:
-          name: installer-amd64
+          name: installer-amd64-built-on-amd64
           path: ${{ github.workspace }}/tmp/amd64
 
       - uses: actions/download-artifact@v4
         with:
-          name: installer-arm64
+          name: installer-arm64-built-on-amd64
```
compnerd, line 840: We shouldn't be changing the name of the artifact, the native vs cross should generate the same content.

.github/workflows/build-toolchain.yml
```diff
@@ -662,17 +827,17 @@ jobs:
 
   release:
     runs-on: ubuntu-latest
-    needs: [context, windows-build]
+    needs: [context, windows-build-x64]
```
compnerd, line 830: This should be dependent on both of the architecture builds.

.github/workflows/build-toolchain.yml
hjyamauchi: Done

.github/workflows/build-toolchain.yml
hjyamauchi: Done

.github/workflows/build-toolchain.yml
hjyamauchi: Ok. `windows-build-arm64` currently fails. Do you know of a way to make it continue on an arm64 failure?

.github/workflows/build-toolchain.yml
hjyamauchi: The problem is that the upload-artifact fails and causes the entire job to fail if something with the same name has already been uploaded and the order between x64/arm64, which depends on the build progress speed, is unpredictable. Can you think of a better workaround?

.github/workflows/build-toolchain.yml
hjyamauchi: Done. But to be tested.

hjyamauchi: This worked.

hjyamauchi:
@compnerd PTAL
A test run (still running): https://github.com/thebrowsercompany/swift-build/actions/runs/11924102225

.github/workflows/build-toolchain.yml
```diff
@@ -284,8 +299,10 @@ jobs:
             fi
           fi
 
-          echo windows_build_runner=${{ inputs.windows_default_runner || vars.WINDOWS_BUILD_RUNNER || 'windows-latest' }} >> ${GITHUB_OUTPUT}
-          echo windows_compilers_runner=${{ inputs.windows_compilers_runner || inputs.windows_default_runner || vars.COMPILERS_BUILD_RUNNER || vars.WINDOWS_BUILD_RUNNER || 'windows-latest' }} >> ${GITHUB_OUTPUT}
+          echo windows_x64_build_runner=${{ inputs.windows_x64_default_runner || vars.WINDOWS_BUILD_RUNNER || 'windows-latest' }} >> ${GITHUB_OUTPUT}
+          echo windows_x64_compilers_runner=${{ inputs.windows_x64_compilers_runner || inputs.windows_x64_default_runner || vars.COMPILERS_BUILD_RUNNER || vars.WINDOWS_BUILD_RUNNER || 'windows-latest' }} >> ${GITHUB_OUTPUT}
+          echo windows_arm64_build_runner=${{ inputs.windows_arm64_default_runner || vars.WINDOWS_BUILD_RUNNER || 'windows-latest' }} >> ${GITHUB_OUTPUT}
+          echo windows_arm64_compilers_runner=${{ inputs.windows_arm64_compilers_runner || inputs.windows_arm64_default_runner || vars.COMPILERS_BUILD_RUNNER || vars.WINDOWS_BUILD_RUNNER || 'windows-latest' }} >> ${GITHUB_OUTPUT}
```
compnerd, line 305: Should we rename the variables to be uniform? We cannot use the same name on both for the native build.

.github/workflows/build-toolchain.yml
```diff
@@ -552,13 +656,13 @@ jobs:
       ANDROID_CMAKE_EXE_LINKER_FLAGS: ${{ needs.context.outputs.ANDROID_CMAKE_EXE_LINKER_FLAGS }}
       ANDROID_CMAKE_SHARED_LINKER_FLAGS: ${{ needs.context.outputs.ANDROID_CMAKE_SHARED_LINKER_FLAGS }}
       ANDROID_NDK_VERSION: ${{ needs.context.outputs.ANDROID_NDK_VERSION }}
-      CMAKE_Swift_FLAGS: ${{ needs.context.outputs.WINDOWS_CMAKE_Swift_FLAGS }}
+      CMAKE_Swift_FLAGS: ${{ needs.context.outputs.CMAKE_Swift_FLAGS }}
```
compnerd, line 659: This undoes some of the changes that @Steelskin was working on I believe.

.github/workflows/build-toolchain.yml
compnerd: That shouldn't matter - we can download only the native builds (i.e. the amd64 installer from x64, the arm64 installer from arm64).

compnerd (commented):
I think that the native only build usage would greatly simplify this change.

hjyamauchi:
> I think that the native only build usage would greatly simplify this change.
What do you mean by "the native only build usage" exactly?

.github/workflows/build-toolchain.yml
compnerd: Right; I'm saying we change that. Only the AMD64 build comes from the AMD64 build and only the ARM64 build comes from the ARM64 build.

.github/workflows/build-toolchain.yml
compnerd: `vars.WINDOWS_BUILD_RUNNER` and `vars.COMPILERS_BUILD_RUNNER` cannot be used simultaneously if you want native compilation. So, should these become `vars.WINDOWS_X64_DEFAULT_BUILD_AGENT` and `vars.WINDOWS_X64_COMPILERS_BUILD_AGENT`, augmented by `vars.WINDOWS_ARM64_{DEFAULT,COMPILERS}_BUILD_AGENT`.

hjyamauchi: Done. Added those variables to `bcny/swift-build`. Do we need to worry about the variables on `compnerd/swift-build`?

.github/workflows/build-toolchain.yml
hjyamauchi: Oh this was accidental. Done.

.github/workflows/build-toolchain.yml
hjyamauchi: @Steelskin This is the discussion we had. It's about how to avoid artifact name conflicts between the cross-compiled arm64 build and the native arm64 build. Currently I use a suffix to avoid collisions. And I use a similar "...built on <os>/<arch>" suffix to the job names to make it less confusing on the GH workflow UI though job name conflicts don't cause a build failure. Appreciate your feedback!

.github/workflows/build-toolchain.yml
```diff
@@ -317,6 +336,8 @@ jobs:
                 },
                 {
                   "arch": "arm64",
+                  "cpu": "aarch64",
```
Steelskin, line 339: IIUC, the intent here is to build the arm64 toolchain from an arm64 host. So the arm64-built-on-x64 and x64-built-on-arm64 jobs should go away since these are building the same artifacts the arm64-built-on-arm64 and x64-built-on-x64 jobs. I believe this would take care of your issue with the naming conflict.

.github/workflows/build-toolchain.yml
```diff
@@ -284,8 +299,10 @@ jobs:
             fi
           fi
 
-          echo windows_build_runner=${{ inputs.windows_default_runner || vars.WINDOWS_BUILD_RUNNER || 'windows-latest' }} >> ${GITHUB_OUTPUT}
-          echo windows_compilers_runner=${{ inputs.windows_compilers_runner || inputs.windows_default_runner || vars.COMPILERS_BUILD_RUNNER || vars.WINDOWS_BUILD_RUNNER || 'windows-latest' }} >> ${GITHUB_OUTPUT}
+          echo windows_x64_build_runner=${{ inputs.windows_x64_default_runner || vars.WINDOWS_BUILD_RUNNER || 'windows-latest' }} >> ${GITHUB_OUTPUT}
+          echo windows_x64_compilers_runner=${{ inputs.windows_x64_compilers_runner || inputs.windows_x64_default_runner || vars.COMPILERS_BUILD_RUNNER || vars.WINDOWS_BUILD_RUNNER || 'windows-latest' }} >> ${GITHUB_OUTPUT}
+          echo windows_arm64_build_runner=${{ inputs.windows_arm64_default_runner || vars.WINDOWS_BUILD_RUNNER || 'windows-latest' }} >> ${GITHUB_OUTPUT}
```
Steelskin, line 304: This is not going to work as it defaults to `windows-latest`, which is an x64 host. Mismatch between the actual bot architecture and the `build_arch` input will cause issues.

.github/workflows/build-toolchain.yml
hjyamauchi: Added `continue-on-error: true` to `windows-build`.

.github/workflows/build-toolchain.yml
Steelskin: I think the main issue is that you are building the same artifacts with both an arm64 and an x64 host (see my other comment). Basically, only one host should be building for one given target. Anything more is redundant.

.github/workflows/build-toolchain.yml
hjyamauchi: ok, based on offline chat, `compnerd/swift-build` does have these variables defined but they are less concern.

.github/workflows/build-toolchain.yml
hjyamauchi: @Steelskin We want to keep three builds: x64-built-on-x64, arm64-built-on-x64 (existing), arm64-built-on-arm64 (new) because the goal here is not just producing the release builds but keep those builds working. Initially we use the first two to produce the release binaries and the third just to ensure that it works.

.github/workflows/build-toolchain.yml
hjyamauchi: Right, agree that's not ideal! What'd happen currently as things are is that it partially works as long as we always pass in `inputs.windows_x64_default_runner` and `inputs.windows_arm64_default_runner`, which I think is currently the case as we override with the cirun machines. If you know of the name to specify the arm64 equivalent of `windows-latest`, lmk. I do not. I don't see a way around it other than requiring the runner parameters in the input and drop the default.

.github/workflows/build-toolchain.yml
hjyamauchi: As above and what compnerd and I discussed in other places, the premise is that we want to keep three builds: x64-built-on-x64, arm64-built-on-x64, arm64-built-on-arm64. They are redundant from the point of view of producing release binaries but they are not from the point of view of ensuring the three builds work.

.github/workflows/build-toolchain.yml
Steelskin: Unfortunately, there is currently no public arm64 windows runner. I believe we should have a toggle for the Windows arm64 build but that definitely complicates things. Let's keep it as-is for now. I'll follow up later when the mac build is in a better state.

.github/workflows/build-toolchain.yml
Steelskin: If we want to keep maintaining the arm64-built-on-x64 build then yes, we don't have a choice, we need the ability to differentiate between the 2 builds or we are going to hit a conflict with the artifact name.

.github/workflows/build-toolchain.yml
hjyamauchi: Gotcha

.github/workflows/build-toolchain.yml
hjyamauchi: Gotcha

hjyamauchi: @compnerd Are you okay with this suffix approach?

.github/workflows/build-toolchain.yml
hjyamauchi: Used `if success() || failure()` on `release` instead of `continue-on-error` as it didn't work.

.github/workflows/swift-toolchain.yml
```diff
@@ -881,17 +877,19 @@ jobs:
           show-progress: false
 
       - name: Install Python ${{ env.PYTHON_VERSION }} (Host)
+        if: matrix.arch == 'amd64' || inputs.build_arch == 'amd64'
```
compnerd, line 880: Why can we not do this on ARM64?

.github/workflows/swift-toolchain.yml
```diff
@@ -881,17 +877,19 @@ jobs:
...
         uses: actions/setup-python@v5
         id: python
         with:
           python-version: '${{ env.PYTHON_VERSION }}'
+          architecture: x64
```
compnerd, line 885: Should this not be `${{ inputs.build_arch }}`?

.github/workflows/swift-toolchain.yml
```diff
@@ -906,19 +904,29 @@ jobs:
...
+          release-asset-name: installer-${{ inputs.build_arch }}.exe
           release-tag-name: '20231016.5'
+      - name: Install Swift Toolchain
+        if: inputs.build_arch == 'arm64'
+        uses: compnerd/gha-setup-swift@b6c5fc1ed5b5439ada8e7661985acb09ad8c3ba2 # main
+        with:
+          github-repo: thebrowsercompany/swift-build
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          release-asset-name: installer-${{ inputs.build_arch }}.exe
+          release-tag-name: '20231016.0'
```
compnerd, line 926: Can we find a single version that works for both and then just use the name to switch between the arches?

.github/workflows/swift-toolchain.yml
```diff
@@ -906,19 +904,29 @@ jobs:
...
 
       - uses: nttld/setup-ndk@v1
+        if: inputs.build_os != 'Windows' || inputs.build_arch != 'arm64'
```
compnerd, line 929: Why can this not run on Windows ARM64?

.github/workflows/swift-toolchain.yml
```diff
@@ -943,8 +951,6 @@ jobs:
           variant: sccache
 
       - name: Configure Compilers
-        env:
```
compnerd, line 946: I dont think that we can remove this

.github/workflows/swift-toolchain.yml
```diff
@@ -1137,15 +1143,19 @@ jobs:
           variant: sccache
 
       - uses: nttld/setup-ndk@v1
+        if: inputs.build_os != 'Windows' || inputs.build_arch != 'arm64'
```
Steelskin, line 1146: Could you factorize this (and others) to this form?
```
!(inputs.build_os == 'Windows' && inputs.build_arch == 'arm64)'
```

It may be just me but I find it easier to infer that we want to run this on everything except for Windows-Arm64 with the negation up front.

compnerd: I don't think that it is just you - that is clearer as to the intent for me as well.

.github/workflows/swift-toolchain.yml
```diff
@@ -340,7 +340,7 @@ jobs:
 
   ds2_tools:
     # TODO: Build this on macOS or make an equivalent Mac-only job
-    if: inputs.build_os == 'Windows'
+    if: inputs.build_os == 'Windows' && inputs.build_arch == 'amd64'
```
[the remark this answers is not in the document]
hjyamauchi: There is no Android NDK for Windows ARM64. So the Android-specific components are omitted from the Windows ARM64 build.

.github/workflows/swift-toolchain.yml
```diff
@@ -389,7 +389,7 @@ jobs:
 
   ds2:
     # TODO: Build this on macOS or make an equivalent Mac-only job
-    if: inputs.build_os == 'Windows'
+    if: inputs.build_os != 'Windows' || inputs.build_arch != 'arm64'
```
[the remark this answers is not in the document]
hjyamauchi: Same

.github/workflows/swift-toolchain.yml
hjyamauchi: I suppose the `actions/setup-python` action doesn't have the right python version for arm64. Hence the use of `nuget` in the arm64 version of this in the next step, as in https://github.com/compnerd/swift-build/pull/390

.github/workflows/swift-toolchain.yml
hjyamauchi: No, as it's used for `amd64` only

.github/workflows/swift-toolchain.yml
hjyamauchi: Here's the context on this: https://thebrowsercompany.slack.com/archives/C07CU2UU97F/p1731980631963529 The `20231016.5` is a specially made build by Alex with a special fix for the main swift-build to work and Alex encountered a build issue for arm64 and hence there's no `20231016.5` for arm64. We cannot simply find a single version because it has a special fix. In addition, the arm64 pinned toolchain needs to be updated anyway as neither `20231016.5` or `20231016.0` or anything close won't work for the correct arm64 build. So, this is a stop-gap placeholder until we have a clearer direction for fixing the arm64 bootstrap process. We could unify when that happens.

.github/workflows/swift-toolchain.yml
hjyamauchi: Because there is no NDK for Windows ARM64

.github/workflows/swift-toolchain.yml
hjyamauchi: Why? I think this env var `NDKPATH` was unused and the build worked without it.

.github/workflows/swift-toolchain.yml
hjyamauchi: Done

.github/workflows/swift-toolchain.yml
```diff
@@ -266,6 +266,10 @@ env:
   WORKAROUND_WINDOWS_PINNED_BOOTSTRAP_TOOLCHAIN_REPO: thebrowsercompany/swift-build
   WORKAROUND_WINDOWS_PINNED_BOOTSTRAP_TOOLCHAIN_RELEASE: 20231016.5
 
+  # Workaround for ... on Windows preventing us from using the 5.10 toolchain release.
```
Steelskin, line 269: I believe @compnerd left the `...` to be filled with an actual reason.

.github/workflows/swift-toolchain.yml
```diff
@@ -1134,15 +1144,19 @@ jobs:
           variant: sccache
 
       - uses: nttld/setup-ndk@v1
+        if: ${{ !(inputs.build_os == 'Windows' && inputs.build_arch == 'arm64') }}
```
Steelskin, line 1147: I think this (and elsewhere) could be simplified with `matrix.os == 'Android'`. I am doing something similar for the Mac build to exclude the NDK flag too. @compnerd WDYT?

.github/workflows/swift-toolchain.yml
```diff
@@ -1668,15 +1709,18 @@ jobs:
                 -D SWIFT_PATH_TO_SWIFT_SYNTAX_SOURCE=${{ github.workspace }}/SourceCache/swift-syntax `
                 -D SWIFT_PATH_TO_STRING_PROCESSING_SOURCE=${{ github.workspace }}/SourceCache/swift-experimental-string-processing
       - name: Build Swift Standard Library
+        if: ${{ !(matrix.os == 'Android' && inputs.build_os == 'Windows' && inputs.build_arch == 'arm64') }}
```
Steelskin, line 1712: Is this still necessary? IIUC, "Android built on Windows Arm64" is disabled with the matrix setup.

.github/workflows/swift-toolchain.yml
hjyamauchi: Done

.github/workflows/swift-toolchain.yml
hjyamauchi: Done.

.github/workflows/swift-toolchain.yml
Steelskin: This is testing if the build is "win/arm64" and the target is "android". I double-checked the matrix definition and I really believe this is never the case.

Steelskin (approved):
lgtm in general. My last concern is that all of the tests for this:
```
if: ${{ !(matrix.os == 'Android' && inputs.build_os == 'Windows' && inputs.build_arch == 'arm64') }}
```

can be removed because they can never be false.

.github/workflows/swift-toolchain.yml
hjyamauchi: Hmm? I see the matrix at lines 1460-1560 above for `stdlib` job, which doesn't seem to restrict Android/Win/ARM64 combination. Can you point me to the matrix definition you're referring to?

Steelskin: Nevermind, my bad. I saw a couple of matrix inputs had been updated and I assumed this one had too.

.github/workflows/swift-toolchain.yml
```diff
@@ -262,10 +262,17 @@ env:
   WORKAROUND_MACOS_PINNED_BOOTSTRAP_TOOLCHAIN_BRANCH: swift-6.0.1-release
   WORKAROUND_MACOS_PINNED_BOOTSTRAP_TOOLCHAIN_TAG: 6.0.1-RELEASE
 
-  # Workaround for ... on Windows preventing us from using the 5.10 toolchain release.
+  # Workaround for the `static_assert(false` failure in Clang on Windows
```
compnerd, line 265:
```suggestion
  # Workaround for the `static_assert(false)` failure in Clang on Windows
```

.github/workflows/swift-toolchain.yml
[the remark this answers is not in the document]
compnerd: Ah, can you add a comment please? I wonder if we can get this to work with the custom clang and their headers and libraries.

.github/workflows/swift-toolchain.yml
compnerd: Ah, okay. I think that because we are doing only native compilation, perhaps we can leave this as `matrix.arch == 'amd64'`?

.github/workflows/swift-toolchain.yml
```diff
@@ -876,17 +880,19 @@ jobs:
...
+        if: matrix.arch == 'amd64' || inputs.build_arch == 'amd64'
         uses: actions/setup-python@v5
         id: python
         with:
           python-version: '${{ env.PYTHON_VERSION }}'
+          architecture: x64
 
       - uses: nuget/setup-nuget@v2
-        if: matrix.arch == 'arm64'
+        if: matrix.arch == 'arm64' || inputs.build_arch == 'arm64'
```
compnerd, line 891: Likewise, because we are only doing native compilation, and this is required for ARM64 only, perhaps we can leave this as `matrix.arch == 'arm64'`.

.github/workflows/swift-toolchain.yml
```diff
@@ -1134,15 +1148,19 @@ jobs:
...
         run: |
-          $NDKPATH = "${{ steps.setup-ndk.outputs.ndk-path }}"
-          if ( "${{ inputs.build_os }}" -eq "Windows" ) {
-            $NDKPATH = cygpath -m $NDKPATH
+          if ("${{ matrix.os }}" -eq "Android") {
+            $NDKPATH = "${{ steps.setup-ndk.outputs.ndk-path }}"
+            if ( "${{ inputs.build_os }}" -eq "Windows" ) {
+              $NDKPATH = cygpath -m $NDKPATH
+            }
+            $CMAKE_NDK_FLAG = "-DCMAKE_ANDROID_NDK=$NDKPATH"
```
compnerd, line 1163: I think that this is why we were configuring the NDK unconditionally - it avoided this logic here. We would always pass along the NDKPATH and then rely on the `CMAKE_SYSTEM_NAME` to deal with the cross-compilation.

.github/workflows/swift-toolchain.yml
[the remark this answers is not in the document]
hjyamauchi: Done

.github/workflows/swift-toolchain.yml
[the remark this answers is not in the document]

.github/workflows/swift-toolchain.yml
hjyamauchi: Done

.github/workflows/swift-toolchain.yml
hjyamauchi: I think we did it unconditionally because we were building only on win/x64 and there's an NDK for it. Since win/arm64 doesn't have one, this logic is necessary because passing an empty `CMAKE_ANDROID_NDK` value causes a failure.

.github/workflows/swift-toolchain.yml
hjyamauchi: I think the second condition is necessary because there's still the matrix case for building win/arm64 on win/x64 (in which case x64 python in the x64 build machine won't be installed) and we need python for both the build and the host arches. I think that's implied by the fact that this step's name has with "(host)" at the end.

.github/workflows/swift-toolchain.yml
```diff
@@ -397,7 +405,8 @@ jobs:
 
   ds2:
     # TODO: Build this on macOS or make an equivalent Mac-only job
-    if: inputs.build_os == 'Windows'
+    # There is no Android NDK for Windows ARM64. So disable on Windows/ARM64.
```
compnerd, line 408: I'm not sure I understand this. What prevents us from shipping DS2 on Windows ARM64? We could use this for remote debugging Windows as well.

.github/workflows/swift-toolchain.yml
compnerd: Oh, I see so that this workflow remains usable for cross-compilation?

.github/workflows/swift-toolchain.yml
compnerd: Might be worth having a comment about that.

.github/workflows/swift-toolchain.yml
```diff
@@ -3669,10 +3741,10 @@ jobs:
...
               -p:INCLUDE_X86_SDK=true `
               -p:INCLUDE_ARM64_SDK=true `
-              -p:ANDROID_INCLUDE_ARM64_SDK=true `
-              -p:ANDROID_INCLUDE_x86_64_SDK=true `
-              -p:ANDROID_INCLUDE_ARM_SDK=true `
-              -p:ANDROID_INCLUDE_X86_SDK=true `
+              -p:ANDROID_INCLUDE_ARM64_SDK=${{ inputs.build_arch == 'amd64' && true || false }} `
+              -p:ANDROID_INCLUDE_x86_64_SDK=${{ inputs.build_arch == 'amd64' && true || false }} `
+              -p:ANDROID_INCLUDE_ARM_SDK=${{ inputs.build_arch == 'amd64' && true || false }} `
+              -p:ANDROID_INCLUDE_X86_SDK=${{ inputs.build_arch == 'amd64' && true || false }} `
```
compnerd, line 3747: Perhaps we should just have a `build_android` setting that is global and use that throughout?

.github/workflows/swift-toolchain.yml
hjyamauchi: Yes!

.github/workflows/swift-toolchain.yml
hjyamauchi: Done

.github/workflows/swift-toolchain.yml
hjyamauchi: I understand ds2 can be used on windows/arm64. But the ds2 job's matrix only contains Android ones (no Windows x64 nor arm64) and ds2_tools are only needed when the ds2 job executes. So I think this seems ok to me. Improved the comment a bit.

.github/workflows/swift-toolchain.yml
hjyamauchi: How do we use a global or workflow-level env variable that depends on `inputs` (like `inputs.build_arch`) here? @compnerd

compnerd: You can define a new parameter for the workflow for that. @Steelskin setup a large number of these. I think for the `inputs` case, we can add a new boolean option.

Steelskin: I believe you should be able to add a `build_android` boolean option to the `swift-toolchain` workflow.
You should be able to infer its value when you call the workflow from `build-toolchain`. It should only be set to true for the Windows sub-workflow, if `needs.context.outputs.windows_build_arch` is `amd64`. Set it to false for the Mac sub-workflow.

compnerd (approved):

Steelskin (approved):

PR #851, opened 2024-11-21
[swift-driver] Simplify the early_swift_driver job

Steelskin:
* Install the early_swift_driver build in `BuildRoot` for consistency.
* Use `gha-upload-tar-artifact` to skip the tar step.
* Remove unneeded actions in the early_swift_driver job.


.github/workflows/swift-toolchain.yml
```diff
@@ -762,25 +762,49 @@ jobs:
...
+
+      - name: Copy binaries
+        run: |
+          if ("${{ matrix.os }}" -eq "Windows") {
+              $ExeSuffix = ".exe"
+              if ("${{ matrix.arch }}" -eq "arm64") {
+                $Cpu = "aarch64"
+              } else {
+                $Cpu = "x86_64"
+              }
```
compnerd, line 785:
```suggestion
              $Cpu = if ("${{ matrix.arch }}" -eq "arm64") { "aarch64" } else { "x86_64" }
```

.github/workflows/swift-toolchain.yml
```diff
@@ -762,25 +762,49 @@ jobs:
...
+              $SourceBinDir = Join-Path "${{ github.workspace }}" "BinaryCache" "swift-driver" "${Cpu}-apple-macosx" "release"
+          }
+          $InstallBinDir = Join-Path "${{ github.workspace }}" "BuildRoot" "bin"
 
-      - id: export-binary-paths
-        run: |
-          $Suffix = if ("${{ matrix.os }}" -eq "Windows") { ".exe" } else { "" }
-          echo "swift_driver=${{ github.workspace }}/BinaryCache/swift-driver/x86_64-unknown-windows-msvc/release/swift-driver${Suffix}" >> $env:GITHUB_OUTPUT
-          echo "swift_help=${{ github.workspace }}/BinaryCache/swift-driver/x86_64-unknown-windows-msvc/release/swift-help${Suffix}" >> $env:GITHUB_OUTPUT
+          # Create the target folder.
```
compnerd, line 794: Why do we need to do this?

Steelskin: The `gha-upload-tar-artifact` action only takes a single path (file or directory). The logic used by the `upload-artifact` and `download-artifact` actions to handle multiple paths and globbing is very hard to replicate.

PR #852, opened 2024-11-21
GHA: enable early swift driver builds on Windows

compnerd:
This wires up the early swift driver build even though this path does not yet work on Windows. This is preliminary support that is waiting on the last changes to enable the early swift driver on Windows.

### DIFF
--- a/.github/workflows/swift-toolchain.yml
+++ b/.github/workflows/swift-toolchain.yml
@@ -711,8 +711,9 @@ jobs:
 
     strategy:
       fail-fast: false
+      matrix: ${{ fromJSON(inputs.host_matrix) }}
 
-    name: Windows amd64 early swift-driver
+    name: ${{ matrix.os }} ${{ matrix.arch }} early swift-driver
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/swift-toolchain.yml
```diff
@@ -747,78 +738,38 @@ jobs:
...
-          New-Item -ItemType Directory -Path $InstallBinDir -Force
+          $env:SWIFTCI_USE_LOCAL_DEPS=1
 
-          # Copy binaries.
-          $SwiftBinaries = @("swift-driver", "swift-help")
-          foreach ($bin in $SwiftBinaries) {
-              $binName = "${bin}${ExeSuffix}"
-              $binPath = Join-Path $InstallBinDir $binName
-              Copy-Item -Path "${SourceBinDir}/${binName}" -Destination $binPath -Force
-          }
```
Steelskin, line 806: We are going to need `gha-upload-tar-artifact` for the Mac build, which only takes a single path or file. I would suggest doing changes here similar to what is done in my [cross PR](https://github.com/compnerd/swift-build/pull/851/files) but with a saner directory name.

compnerd: Let's get this building on Windows first. Once we have that, we can address the use of gha-upload-tar-artifact.

.github/workflows/swift-toolchain.yml
```diff
@@ -707,28 +707,25 @@ jobs:
...
-    # TODO: Build this on Windows too.
-    if: inputs.build_os == 'Darwin'
-    needs: [sqlite]
     runs-on: ${{ inputs.default_build_runner }}
 
     strategy:
       fail-fast: false
-      matrix: ${{ fromJSON(inputs.host_matrix) }}
 
-    name: ${{ matrix.os }} ${{ matrix.arch }} Early Swift Driver
```
Steelskin, line 719: This already builds on macOS. Could you change the matrix to `build_matrix` if this is what we need?

.github/workflows/swift-toolchain.yml
```diff
@@ -755,78 +753,39 @@ jobs:
...
-          foreach ($bin in $SwiftBinaries) {
-              $binName = "${bin}${ExeSuffix}"
-              $binPath = Join-Path $InstallBinDir $binName
-              Copy-Item -Path "${SourceBinDir}/${binName}" -Destination $binPath -Force
-          }
+          swift build `
+            --configuration release `
+            --package-path ${{ github.workspace }}/SourceCache/swift-driver `
+            --build-path ${{ github.workspace }}/BinaryCache/swift-driver `
+            -Xlinker "${env:SDKROOT}/usr/lib/swift/windows/x86_64/swiftCore.lib"
```
compnerd, line 770: This won't work on macOS and will need to be figured out.

.github/workflows/swift-toolchain.yml
```diff
@@ -755,78 +753,39 @@ jobs:
...
 
-          # Create an archive to preserve permissions.
-          tar -czf "${{ github.workspace }}/BinaryCache/early-swift-driver.tar.gz" `
-              -C "${{ github.workspace }}/BinaryCache" `
-              $InstallFolderName
+      - id: export-binary-paths
+        run: |
+          $Suffix = if ("${{ matrix.os }}" -eq "Windows") { ".exe" } else { "" }
+          echo "swift_driver=${{ github.workspace }}/BinaryCache/swift-driver/x86_64-unknown-windows-msvc/release/swift-driver${Suffix}" >> $env:GITHUB_OUTPUT
+          echo "swift_help=${{ github.workspace }}/BinaryCache/swift-driver/x86_64-unknown-windows-msvc/release/swift-help${Suffix}" >> $env:GITHUB_OUTPUT
```
compnerd, line 776: These won't work on macOS and need to be figured out.

.github/workflows/swift-toolchain.yml
Steelskin: Ack, I'll fix it.

Steelskin (approved):

PR #853, opened 2024-11-26
Refactor matrix definitions

Steelskin:
The various build matrix definitions used to be duplicated across multiple jobs. This changes the matrix definitions to have a single definition for every OS-Arch target, removing duplication. In addition, this change makes it easier to set up a separate arm64 Windows build.

This PR breaks the Mac build. It will be fixed as all jobs will be updated to use the new setup. Subsequent PRs will introduce Windows-only and Android-only matrices passed to the sub-build.

The end goal is to have a single matrix definition for every platform and remove many of the inputs to the `swift-toolchain` sub-build. This also allows for re-merging the Mac and Windows sub-builds, reducing build complexity.

### DIFF
--- a/.github/workflows/build-toolchain.yml
+++ b/.github/workflows/build-toolchain.yml
@@ -81,6 +81,31 @@ on:
         type: string
         required: false
 
+      build_setup:
+        description: >-
+          JSON definition of the build setup. This is a dictionary in the 
+          following format:
+          ```
+          {
+            "builder1": [ "target1", "target2" ],
+            "builder2": [ "target3", "target4" ]
+          }
+          ```
+          The valid `builder` values are:
+            - Windows-amd64
+            - Darwin
+          The valid `target` values are:
+            - Windows-amd64
+            - Windows-arm64
+            - Android
+            - Darwin
+          Note that not every combination will result in a successful build. 
+          The workflow will fail early if invalid values are provided. The 
+          default setup is specified by the `DEFAULT_BUILD_SETUP` environment 
+          variable, defined below.
+        type: string
+        required: false
+
     secrets:
       SYMBOL_SERVER_PAT:
         required: true
@@ -91,6 +116,10 @@ on:
 
 env:
   SCCACHE_DIRECT: yes
+  DEFAULT_BUILD_SETUP: >-
+    {
+      "Windows-amd64": [ "Windows-amd64", "Windows-arm64", "Android" ]
+    }
 
 jobs:
   context:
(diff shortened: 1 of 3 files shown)


Steelskin:
PTAL. I'd like to find a better place for `setup-matrix.ps1`. We've been putting these scripts directly in the build file but that makes it really hard to test locally. Please, let me know what you think.

.github/workflows/build-toolchain.yml
```diff
@@ -301,208 +329,191 @@ jobs:
...
+              "cpu": "aarch64",
+              "os": "Android",
+              "platform": "Android",
+              "triple": "aarch64-unknown-linux-android${{ steps.context.outputs.ANDROID_API_LEVEL }}",
+              "triple_no_api_level": "aarch64-unknown-linux-android",
+              "cc": "clang",
+              "cflags": "${{ steps.context.outputs.ANDROID_CMAKE_C_FLAGS }}",
+              "cxx": "clang++",
+              "cxxflags": "${{ steps.context.outputs.ANDROID_CMAKE_CXX_FLAGS }}",
+              "swiftflags": "-sdk $NDKPATH/toolchains/llvm/prebuilt/windows-x86_64/sysroot -Xclang-linker -target -Xclang-linker aarch64-unknown-linux-android${{ steps.context.outputs.ANDROID_API_LEVEL }} -Xclang-linker --sysroot -Xclang-linker $NDKPATH/toolchains/llvm/prebuilt/windows-x86_64/sysroot -Xclang-linker -resource-dir -Xclang-linker $NDKPATH/toolchains/llvm/prebuilt/windows-x86_64/lib/clang/17 -L ${GithubWorkspace}/BinaryCache/swift/lib/swift/android -g",
```
hjyamauchi, line 395: I think a potential issue might be with declaring all the flags here be that there are may be flags that need to be set conditionally. For example, there's no NDK for Windows/arm64. We'd need to elide the `NDKPATH` from the flags conditionally if it's for Windows/arm64 and we need drop the flags like `-DCMAKE_ANDROID_NDK` and `-DSWIFT_ANDROID_NDK_PATH` instead of passing empty values, or they will err. For example, around `swift-toolchain.yml:1659` in [the PR](https://github.com/compnerd/swift-build/pull/847/files). Could we avoid that, as in would it be a real issue?

setup-matrix.ps1
```diff
@@ -0,0 +1,170 @@
...
+.PARAMETER BuildSetup
+A JSON string that contains the setup for the matrices. The JSON should be an
+object with keys as the builders names and values as an array of target names.
+
+.EXAMPLE
+./setup-matrix.ps1 -BuildSetup '{
+    "Windows-amd64": ["Windows-amd64"],
+    "Darwin": ["Darwin"]
+}'
+#>
```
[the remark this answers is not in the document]
Steelskin: Thanks for pointing this out, there was actually a bug in the host matrix definition, which I fixed.

.github/workflows/build-toolchain.yml
Steelskin: When I was working on getting the `sdk` job to build on Mac (which I have since abandoned), I encountered a similar issue. I solved it by making the NDK flags conditional to `"${{ matrix.os }}" -eq "Android"` in a preliminary `setup-config` step. See this [branch](https://github.com/compnerd/swift-build/commit/e5bd0a6b0cf941a93d7be2fc69379683fb40d094#diff-cd548d9f13bbeda6b71038430c870dd197f12e4006bdfdb5f87c6c7f442caf81R2008)

This should not be a problem if I can adapt this WIP branch on top of this PR to only add the NDK-related arguments for Android. This will also not be part of any build matrix with Windows Arm64 as the builder. Does that make sense?

hjyamauchi: It sounds like there's a way to handle it, sure :)

hjyamauchi (approved):

Steelskin:
Abandoning this change, as discussed offline.

PR #857, opened 2024-12-04
[compilers] Add libxml2 to the compilers job

Steelskin:
libxml2 was added as a required dependency downstream in swiftlang/swift#77862. This brings in the libxml2 build to the compilers job to fix the build.


hjyamauchi (approved):

compnerd (approved):
Sure, this is fine to unbreak the build

PR #858, opened 2024-12-04
GHA: adjust CMake invocation for libxml2 usage

compnerd:
Use the CMake CONFIG definitions to wire up LibXml2 rather than trying to manually add the dependencies.


Steelskin (approved):
Thanks, I was just doing the exact same change.

PR #862, opened 2024-12-09
GHA: correct the libxml2 path for `LibXml2_DIR`

compnerd:
There is no architecture component in the path with the YAML workflow.


Steelskin (approved):
Ah, missed that one.

PR #865, opened 2024-12-12
[Mac build] Update upload/download actions

Steelskin:
This updates the upload and download actions to use the tar-artifact actions for the compilers and cmark-gfm artifacts, paving the way for the Mac compilers build.


Steelskin:
Here is the corresponding run: https://github.com/thebrowsercompany/swift-build/actions/runs/12301415539

PR #871, opened 2024-12-16
GHA: use `llvm-mt` instead of `mt`

compnerd:
Now that we build with libxml2, use the LLVM manifest tool over the Microsoft one.


hjyamauchi (approved):

compnerd:
CI run at https://github.com/thebrowsercompany/swift-build/actions/runs/12361117119

PR #876, opened 2024-12-18
[devtools] Add swift-lmdb to the devtools job

Steelskin:
swift-lmdb is now a required dependency of IndexStoreDB.


compnerd (commented):
You need to pipe that into DocC-SymbolKit and SourceKit-LSP as well.

Steelskin (commented):
@compnerd PTAL
As I mentioned, I had to add this to a couple of build tasks to workaround a linker bug:
```
-Xfrontend -disable-round-trip-debug-type
```

Let me know if you'd rather try fixing that problem separately. I got a green build with this PR.

Steelskin:
> Yeah, I think that we should try to debug that problem separately. If we can identify where it got introduced, we should be able to revert it. But the fact that we are building this in CI and not hitting it there is suspicious.

Reverted the flags. I think I can narrow the issue down to a small range to debug further.

compnerd (approved):

PR #879, opened 2025-01-13
[Mac build] Add target the Mac host matrix

Steelskin:
`compiler_target` and `cpu` are required for the early swift driver build since #878.


Steelskin:
Test run is here: https://github.com/thebrowsercompany/swift-build/actions/runs/12751395362/job/35538507108 (still running on Windows but the Mac side is fixed)

hjyamauchi (approved):

PR #880, opened 2025-01-14
GHA: emit a ToolchainInfo.plist

compnerd:
This was introduced into the upstream toolchain build, replicate it here to ensure that we are able to package the toolchain.


.github/workflows/swift-toolchain.yml
```diff
@@ -1151,6 +1151,19 @@ jobs:
             Copy-Item -Path "${{ github.workspace }}/BinaryCache/Library/cmark-gfm-${{ inputs.swift_cmark_version }}/usr/lib/*.dylib" -Destination "${{ github.workspace }}/BuildRoot/Library/Developer/Toolchains/unknown-Asserts-development.xctoolchain/usr/lib/swift/host/compiler"
           }
 
+      - uses: actions/setup-python@v5
+      - uses: jannekem/run-python-script-action@v1
```
Steelskin, line 1155: Why not use powershell for this?

hjyamauchi (approved):

.github/workflows/swift-toolchain.yml
Steelskin: I suppose not. I was indeed thinking about generating the XML using Powershell. I understand if you'd rather not, though. LGTM

Steelskin (approved):

PR #882, opened 2025-01-22
GHA: remove Yams from the build

compnerd:
This is no longer used as a dependency for swift-driver, remove the unnecessary build.


compnerd:
Triggered Job: https://github.com/thebrowsercompany/swift-build/actions/runs/12918108043

hjyamauchi (approved):

PR #883, opened 2025-01-23
[packaging] Update value for ProductArchitecture

Steelskin:
In swiftlang/swift-installer-scripts#358, the ProductArchitecture value for packaging the Android SDK was changed.

### DIFF
--- a/.github/workflows/swift-toolchain.yml
+++ b/.github/workflows/swift-toolchain.yml
@@ -3680,6 +3680,12 @@ jobs:
       - name: Package SDK
         if: inputs.build_android
         run: |
+          if ("${{ inputs.build_arch }}" -eq "amd64") {
+            $InstallerPlatform = "x64"
+          } else {
+            $InstallerPlatform = "arm64"
+          }
+
           msbuild -nologo -restore -maxCpuCount `
               -p:BaseOutputPath=${{ github.workspace }}\BinaryCache\installer\ `
               -p:Configuration=Release `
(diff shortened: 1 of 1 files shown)


.github/workflows/swift-toolchain.yml
```diff
@@ -3613,15 +3613,19 @@ jobs:
         include:
           - arch: arm64
             cpu: aarch64
+            product_arch: arm64
```
compnerd, line 3616: `msarch` might be better?

.github/workflows/swift-toolchain.yml
```diff
@@ -3676,6 +3680,12 @@ jobs:
       - name: Package SDK
         if: inputs.build_android
         run: |
+          if ("${{ inputs.build_arch }}" -eq "amd64") {
```
compnerd, line 3683: Is this supposed to be `build_arch` or `host_arch`?

compnerd (approved):

.github/workflows/swift-toolchain.yml
compnerd: To connect the dots ... the reason that we should then use the build architecture is that we are now doing native compilation only after @hjyamauchi's changes?

Steelskin: I believe these packages used to be built for the `x86` platform. Let's follow up next week. These changes - along with the `swift-installer-scripts` fix the build for now.

PR #888, opened 2025-01-30
[stdlib] Change paths for stdlib job and artifacts

Steelskin:
Currently, in the `stdlib` job, the toolchain is extracted to `BuildRoot/Library`, and the Swift Standard Library is also installed to `BuildRoot/Library`. Other jobs typically extract the toolchain to `BinaryCache/Library`. This changes the extraction path to `BinaryCache/Library` in the `stdlib` job, bringing it in-line with the rest of the build.

In addition, this changes the Swift Standard Library artifact root to be `BuildRoot/Library`, rather than the full path to the platform SDK, `BuildRoot/Library/Developer/Platforms/${{ matrix.os }}.platform`. This is in preparation for adapting the job to build for macOS, where the Swift Standard Library will be installed under `Library/Toolchains/unknown-Asserts-development.xctoolchain/usr`.

Extracted from #843

### DIFF
--- a/.github/workflows/swift-toolchain.yml
+++ b/.github/workflows/swift-toolchain.yml
@@ -1690,7 +1690,7 @@ jobs:
         uses: thebrowsercompany/gha-download-tar-artifact@59992d91335d4ecba543c8535f7d07238e42125d # main
         with:
           name: compilers-Windows-${{ inputs.build_arch }}
-          path: ${{ github.workspace }}/BuildRoot/Library
+          path: ${{ github.workspace }}/BinaryCache/Library
       - uses: actions/checkout@v4
         with:
           repository: swiftlang/llvm-project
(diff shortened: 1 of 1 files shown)


Steelskin:
PTAL. Here's the test job: https://github.com/thebrowsercompany/swift-build/actions/runs/13060564334
It will make #843 easier to parse.

compnerd (commented):
There are two changes here:
1. rooting the artifacts higher to allow us to avoid having to specify the intermediary components
2. Changing the location that the toolchain is extracted to

1 is clearly a simplification with no impact.

2 however is interesting. The differentiation between `BinaryCache` and `BinaryRoot` is that the `BinaryCache` is the current job's build artifacts (it is effectively a cache of the output). The `BinaryRoot` is the partially staged toolchain that we are building up. By separating the two it makes it easier to overlap the content as we build up more components. This would make that more difficult. This is a trade off, and I was wondering if you could go into more details of how you see this simplifying the build.

Steelskin:
@compnerd Following our offline discussion, I updated the commit message to clarify.

compnerd (approved):
Please squash and merge!

PR #892, opened 2025-02-04
Update dependencies to match the swift repository

Steelskin:


compnerd (approved):
I wonder if we can remove the default values and just use the repo manifest to compute the revisions for the non-snapshot case.

PR #895, opened 2025-02-06
[sdk] Add "-Xlinker" to linker for Swift projects

Steelskin:
Since CMake 3.30, `CMAKE_[*]_LINKER_FLAGS` are passed to the linker invocation as-is. This causes issues in Swift-only projects where the linker invocation is `swiftc` used as a driver for the actual underlying linker. `swiftc` cannot parse these arguments and fails. As a workaround, this rewrites the `shared_linker_flags` and `exe_linker_flags`, adding `-Xlinker` for each flag, which solves the problem.
In the future, CMake 4.0 introduces `CMP0181` which will allow `CMAKE_[*]_LINKER_FLAGS` to interpret the `LINKER:` prefix and generate the correct linker invocation.


.github/workflows/swift-toolchain.yml
```diff
@@ -2282,6 +2285,13 @@ jobs:
           $CLANG_CL = cygpath -m ${{ github.workspace }}/BinaryCache/Library/Developer/Toolchains/unknown-Asserts-development.xctoolchain/usr/bin/clang-cl.exe
           $SWIFTC = cygpath -m ${{ github.workspace }}/BinaryCache/Library/Developer/Toolchains/unknown-Asserts-development.xctoolchain/usr/bin/swiftc.exe
 
+          # Worarkound CMake 3.30 issue where CMAKE_[*]_FLAGS are passed as-is to the linker driver.
+          # TODO: Once we have CMake 4.0, set CMP0181 to NEW and pass these as "LINKER:" flags.
+          $CMAKE_SHARED_LINKER_FLAGS =
+            ("${{ matrix.shared_linker_flags }}".Split(" ").Where({ $_.Trim() -ne ''}) | ForEach-Object { "-Xlinker $_" }) -join " "
```
compnerd, line 2291: What happens if there is a space in a path?

Steelskin: It will break, but it also breaks currently too as CMake also parses the space as a separator. Note that we don't perform any kind of escape for quotes here either.
The good news is that when we switch to CMake 4.0, we'll be able to handle these properly since `LINKER:` supports comma-separated values.

compnerd: Hmm, I thought it was possible to get that passed properly. Well, I guess if it's no worse off ...

compnerd (approved):

PR #905, opened 2025-02-26
GHA: adjust the file system layout to accommodate the installer

compnerd:
This adjusts the installation to shuffle around some of the installed files to match the expected layout post-installation. This is required to accommodate the installer which is beginning to expect the toolchain to be laid out as desired.

### DIFF
--- a/.github/workflows/swift-toolchain.yml
+++ b/.github/workflows/swift-toolchain.yml
@@ -2366,6 +2366,13 @@ jobs:
 
           $CMAKE_CPU = if ("${{ matrix.cpu }}" -eq "armv7") { "armv7-a" } else { "${{ matrix.cpu }}" }
 
+          $XCTestBinDir = switch ("${{ matrix.cpu }}") {
+          "armv7" { "bin32a" }
+          "i686" { "bin32" }
+          "x86_64" { "bin64" }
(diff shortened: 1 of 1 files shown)


compnerd:
Test run at https://github.com/thebrowsercompany/swift-build/actions/runs/13551049983

.github/workflows/swift-toolchain.yml
```diff
@@ -2366,6 +2366,13 @@ jobs:
 
           $CMAKE_CPU = if ("${{ matrix.cpu }}" -eq "armv7") { "armv7-a" } else { "${{ matrix.cpu }}" }
 
+          $XCTestBinDir = switch ("${{ matrix.cpu }}") {
+          "armv7" { "bin32a" }
```
hjyamauchi, line 2370: (Maybe we want an indentation for the switch cases?)

hjyamauchi (approved):

.github/workflows/swift-toolchain.yml
compnerd: 🤷‍♂️ I'm fine with that being adjusted in the future. LLVM doesn't indent the case statements IIRC, so I didn't do that here. Let's get this merged to unblock CI.

PR #906, opened 2025-02-26
Skip more steps if Android is disabled

Steelskin:
This speeds up the no-Android builds by disabling unneeded downloads and repo syncs.


compnerd (approved):
This is nice. I wonder if we can hoist the Android check into a variable computed once for readability.

hjyamauchi (approved):

PR #908, opened 2025-02-27
GHA: allow overriding the repo manifest

compnerd:
Enable building with a custom fork of the repo manifest enabling patched forks to be used as the source of a build.

### DIFF
--- a/.github/workflows/build-toolchain.yml
+++ b/.github/workflows/build-toolchain.yml
@@ -3,6 +3,12 @@ name: Development Snapshot
 on:
   workflow_dispatch:
     inputs:
+      repo_url:
+        description: 'Repo Manifest URL'
+        default: 'https://github.com/compnerd/swift-build'
+        required: false
+        type: string
+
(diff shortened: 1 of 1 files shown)


hjyamauchi (approved):
Nice!

hjyamauchi:
Sorry for a late comment. I think we'd want to parameterize the `branch_name` as well so we can point to not only a person's branch but also a branch in it

compnerd:
> Sorry for a late comment. I think we'd want to parameterize the `branch_name` as well so we can point to not only a person's branch but also a branch in it

That is not possible - we no longer have any free inputs left. The workflow dispatch is limited to 10 inputs.

hjyamauchi:
ok does that mean that the branch name of a person's repo must be either main or a release version number like 6.0.1, I think

compnerd:
> ok does that mean that the branch name of a person's repo must be either main or a release version number like 6.0.1, I think

Yes, that is the restriction.

PR #914, opened 2025-03-07
Fix up Testing/XCTest/DS2 install paths

hjyamauchi:


.github/workflows/swift-toolchain.yml
```diff
@@ -427,6 +427,7 @@ jobs:
             cxxflags: ${{ inputs.ANDROID_CMAKE_CXX_FLAGS }}
             os: Android
             extra_flags: -DCMAKE_ANDROID_API=${{ inputs.ANDROID_API_LEVEL }} -DCMAKE_ANDROID_ARCH_ABI=arm64-v8a
+            triple_no_api_level: aarch64-unknown-linux-android
```
compnerd, line 430: Can we use `module_triple` for the name?

.github/workflows/swift-toolchain.yml
```diff
@@ -518,6 +522,11 @@ jobs:
         if: inputs.build_android
         run: cmake --build ${{ github.workspace }}/BinaryCache/ds2 --target install
 
+      - name: DS2 Install Fixup
+        if: inputs.build_android
+        run: |
+          Move-Item -Path "${{ github.workspace }}/BinaryCache/Library/Developer/ds2/usr/bin/ds2" "${{ github.workspace }}/BinaryCache/Library/Developer/ds2/usr/bin/${{ matrix.triple_no_api_level }}-ds2"
```
compnerd, line 528: I think that none of this should be needed and we can just use `DS2_PROGRAM_PREFIX=${{ matrix.module_triple }}` if Android.

.github/workflows/swift-toolchain.yml
hjyamauchi: Done

.github/workflows/swift-toolchain.yml

compnerd (approved):

PR #918, opened 2025-03-11
GHA: shuffle the Foundation headers

compnerd:
The installer rules are progressively assuming a fully staged toolchain. We did not previously shuffle the C headers that Foundation expects into the correct location. Do so to allow packaging with the updated manifest.


compnerd:
Tested at https://github.com/thebrowsercompany/swift-build/actions/runs/13910825500

hjyamauchi (approved):

PR #924, opened 2025-03-21
GHA: adjust the XCTest and Testing CMake invocations

compnerd:
Bring this in line with the build.ps1 equivalent, removing the file shuffling post-installation.


compnerd:
Tested at https://github.com/thebrowsercompany/swift-build/actions/runs/13981078011

PR #930, opened 2025-03-28
Add dynamic configuration of sccache based on available environment variables

brianmichel:
We want to dynamically upgrade the sccache configuration if S3 access via SSO is setup. To do this while preserving backward disk caching compatibility we specify 3 new environment variables:
- `SCCACHE_S3_BUCKET`
- `SCCACHE_AWS_REGION `
- `SCCACHE_AWS_ARN `

If the `SCCACHE_AWS_ARN` is set the sccache action will attempt to validate that the other two values are not null or an empty string before pushing them into the environment for sccache to pick up.

We have to also specify `id-token: write` permissions at the top level entry points into the CI process since github tokens cannot be upgraded when workflows get called (they can only get downgraded) so we have to put the complete set of permissions at the root of CI.

### DIFF
--- a/.github/actions/setup-sccache/action.yml
+++ b/.github/actions/setup-sccache/action.yml
@@ -43,12 +43,12 @@ runs:
           }
 
           Write-Host "Using S3 bucket ${{ inputs.s3-bucket }} for cache storage."
-          Write-Host SCCACHE_BUCKET=${{ inputs.s3-bucket }} | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
-          Write-Host SCCACHE_REGION=${{ inputs.aws-region }} | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
-          Write-Host SCCACHE_S3_SERVER_SIDE_ENCRYPTION=${{ inputs.s3-bucket-encryption }} | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
+          "SCCACHE_BUCKET=${{ inputs.s3-bucket }}" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
+          "SCCACHE_REGION=${{ inputs.aws-region }}" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
+          "SCCACHE_S3_SERVER_SIDE_ENCRYPTION=${{ inputs.s3-bucket-encryption }}" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
         } else {
           Write-Host "Using local disk cache."
-          Write-Host SCCACHE_DIRECT='on' | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
+          "SCCACHE_DIRECT=on" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
         }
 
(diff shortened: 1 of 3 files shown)


.github/actions/setup-sccache/action.yml
```diff
@@ -43,12 +43,12 @@ runs:
...
-          Write-Host SCCACHE_BUCKET=${{ inputs.s3-bucket }} | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
-          Write-Host SCCACHE_REGION=${{ inputs.aws-region }} | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
-          Write-Host SCCACHE_S3_SERVER_SIDE_ENCRYPTION=${{ inputs.s3-bucket-encryption }} | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
+          "SCCACHE_BUCKET=${{ inputs.s3-bucket }}" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
+          "SCCACHE_REGION=${{ inputs.aws-region }}" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
+          "SCCACHE_S3_SERVER_SIDE_ENCRYPTION=${{ inputs.s3-bucket-encryption }}" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
         } else {
           Write-Host "Using local disk cache."
-          Write-Host SCCACHE_DIRECT='on' | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
+          "SCCACHE_DIRECT=on" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
```
brianmichel, line 51: I had to switch these back since `Write-Host` was just printing them to the console output and not into the env for sccache to pick up in the later step.

.github/workflows/build-toolchain.yml
```diff
@@ -788,6 +786,9 @@ jobs:
   windows-build:
     needs: [context]
     name: Windows Swift Toolchains Build
+    permissions:
+      contents: read
+      id-token: write
```
compnerd, line 791: Why the change to the permissions?

brianmichel: we need to specify `id-token` here since that's what is used to allow the OIDC connection between GitHub and AWS to work (if someone is specifying the AWS_ARN). `contents: read` is a standard permission for being able to read content from the repo itself, like listing commits. I can try to remove the `contents` one and see what happens, but we have to specify `id-token` since the built in github token that is used for workflows can only ever be _downgraded_ in scope, not upgraded.

Since we use a bunch of callable workflows we have to specify all of the permissions that any workflow that might be called could need (in our case we're now saying the need to be able to potentially generate OIDC tokens). Let me know if that makes sense, happy to tuple about it.

Other reading: https://docs.github.com/en/actions/sharing-automations/reusing-workflows#example-caller-workflow:~:text=The%20GITHUB_TOKEN%20permissions%20passed%20from%20the%20caller%20workflow%20can%20be%20only%20downgraded%20(not%20elevated)%20by%20the%20called%20workflow.

compnerd: Okay; just was trying to understand why the extra permissions were being added - I like the default behaviour of only allowing downgrading the scopes rather than upgrading.

compnerd (approved):

PR #931, opened 2025-04-02
Create vfs-overlay.yaml on demand

Steelskin:
`vfs-overlay.yaml` created as part of the `stdlib` step refers to paths from the `swift` source directory, which may not be available in subsequent jobs, causing build issues. These changes generate a `vfs-overlay.yaml` file on demand for jobs that need it, to work around the issue.

### DIFF
--- a/.github/workflows/swift-toolchain.yml
+++ b/.github/workflows/swift-toolchain.yml
@@ -2016,7 +2016,7 @@ jobs:
                   type: file
                   external-contents: "${ModuleMapDir}/vcruntime.apinotes"
           "@
-          $VfsOverlayDir = Split-Path -parent $VfsOverlay
+          $VfsOverlayDir = Split-Path -Parent $VfsOverlay
           New-Item -ItemType Directory -Path $VfsOverlayDir
           Write-Output $ModuleMap | Out-File -FilePath $VfsOverlay -Encoding utf8
 
@@ -2380,7 +2380,7 @@ jobs:
                   external-contents: "${ModuleMapDir}/vcruntime.apinotes"
           "@
 
-          $VfsOverlayDir = Split-Path -parent $VfsOverlay
+          $VfsOverlayDir = Split-Path -Parent $VfsOverlay
           New-Item -ItemType Directory -Path $VfsOverlayDir
           Write-Output $ModuleMap | Out-File -FilePath $VfsOverlay -Encoding utf8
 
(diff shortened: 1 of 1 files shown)


Steelskin:
Try job: https://github.com/thebrowsercompany/swift-build/actions/runs/14223822621

compnerd:
I'm not sure that I understand the need for this. Why not simply create the VFS Overlay from scratch if needed?

.github/workflows/swift-toolchain.yml
```diff
@@ -1987,15 +1970,55 @@ jobs:
...
+      - name: Create vfs-overlay
+        run: |
           $VfsOverlay = "${{ github.workspace }}/BinaryCache/swift/stdlib/windows-vfs-overlay.yaml"
           $ModuleMapDir = cygpath -m ${{ github.workspace }}/BinaryCache/Library/Developer/Platforms/Windows.platform/Developer/SDKs/Windows.sdk/usr/share
-          (Get-Content $VfsOverlay).Replace("<MODULE_MAP_DIR>", "${ModuleMapDir}") | Set-Content $VfsOverlay
+          $Win10SdkRoot = Get-ItemPropertyValue `
+            -Path "Registry::HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows Kits\Installed Roots" `
+            -Name "KitsRoot10"
+          $Win10SdkRoot = cygpath -m ${Win10SdkRoot}
+          $InstallerLocation = Join-Path "${env:ProgramFiles(x86)}" "Microsoft Visual Studio" "Installer"
```
compnerd, line 1987: I think that [IO.Path]::Combine is slightly better, but let’s get @jeffdav’s take on it.

compnerd (approved):
Other than what @jeffdav says about the path handling, LGTM

.github/workflows/swift-toolchain.yml
```diff
@@ -1980,12 +1970,56 @@ jobs:
...
+              type: directory
+              contents:
+                - name: module.modulemap
+                  type: file
+                  external-contents: "${ModuleMapDir}/vcruntime.modulemap"
+                - name: vcruntime.apinotes
+                  type: file
+                  external-contents: "${ModuleMapDir}/vcruntime.apinotes"
+          "@
+          $VfsOverlayDir = Split-Path -parent $VfsOverlay
```
jeffdav, line 2019: nit.
```suggestion
          $VfsOverlayDir = Split-Path -Parent $VfsOverlay
```

jeffdav (approved):
Double check that path thing...

.github/workflows/swift-toolchain.yml
Steelskin: This works with PowerShell Core (pwsh), which is what we use on CI.

PR #932, opened 2025-04-04
Setup remote cache after auth success

brianmichel:
It seems like there is some eagerness with GitHub actiosn starting to run steps after a previous step is not totally complete or some issue with ensuring those env variables are set correctlly across steps. This forces us to wait until after we've successfully assumed the credentials role to configure the cache.


brianmichel:
Fixes issue https://linear.app/the-browser-company/issue/DEVEX-2805/sccache-setup-is-flakey

Steelskin (approved):

PR #935, opened 2025-04-07
GHA: shuffle more files for Dispatch, Foundation

compnerd:
The toolchain build outside of libdispatch and Foundation now properly install into the architecture specific locations. The installer rules now expect the properly staged toolchain image rather than try to adjust the location. This shuffling ensures that the location of the import libraries matches the expectations.


Steelskin (approved):

compnerd:
Going to merge this for now, it is speculative, but should be correct. The NoAsserts builds are causing some trouble atm.

PR #941, opened 2025-04-21
Packaging: Set `productArchitecture`

Steelskin:
This fixes the Windows Arm64 build job.

### DIFF
--- a/.github/workflows/swift-toolchain.yml
+++ b/.github/workflows/swift-toolchain.yml
@@ -3815,6 +3815,7 @@ jobs:
               -p:ImageRoot=${{ github.workspace }}/BuildRoot/Library/Developer `
               -p:WORKAROUND_MIMALLOC_ISSUE_997=$WORKAROUND_MIMALLOC_ISSUE_997 `
               -p:ProductVersion=${{ inputs.swift_version }} `
+              -p:ProductArchitecture=${{ matrix.arch }} `
               ${{ github.workspace }}/SourceCache/swift-installer-scripts/platforms/Windows/bld/bld.wixproj
 
       - name: Package CLI Tools
--- a/.github/workflows/swift-toolchain.yml
+++ b/.github/workflows/swift-toolchain.yml
@@ -3828,6 +3828,7 @@ jobs:
               -p:PASSPHRASE=${{ secrets.PASSPHRASE }} `
               -p:ImageRoot=${{ github.workspace }}/BuildRoot/Library/Developer `
               -p:ProductVersion=${{ inputs.swift_version }} `
+              -p:ProductArchitecture=${{ matrix.arch }} `
               ${{ github.workspace }}/SourceCache/swift-installer-scripts/platforms/Windows/cli/cli.wixproj
 
       - name: Package Debugging Tools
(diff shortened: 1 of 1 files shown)


Steelskin:
Try job: https://github.com/thebrowsercompany/swift-build/actions/runs/14574536193

compnerd (commented):
Should we not be passing that across the other MSIs as well?

Steelskin:
> Should we not be passing that across the other MSIs as well?

It seems to be only used by that one target but I don't know if msbuild and/or wix also use that property internally.

Steelskin:
> > Should we not be passing that across the other MSIs as well?
> 
> It seems to be only used by that one target but I don't know if msbuild and/or wix also use that property internally.

Actually I was wrong, I updated the PR. Here's the new try job: https://github.com/thebrowsercompany/swift-build/actions/runs/14577741095

.github/workflows/swift-toolchain.yml
```diff
@@ -3874,26 +3878,26 @@ jobs:
         with:
           name: Windows-${{ matrix.arch }}-bld-msi
           path: |
-            ${{ github.workspace }}/BinaryCache/installer/Release/${{ inputs.build_arch }}/bld.msi
-            ${{ github.workspace }}/BinaryCache/installer/Release/${{ inputs.build_arch }}/bld.cab
+            ${{ github.workspace }}/BinaryCache/installer/Release/${{ matrix.arch }}/bld.msi
+            ${{ github.workspace }}/BinaryCache/installer/Release/${{ matrix.arch }}/bld.cab
```
compnerd, line 3882: Hmm, I had to change this to `inputs.build_arch`. Does this work?

.github/workflows/swift-toolchain.yml
```diff
@@ -4066,19 +4071,19 @@ jobs:
...
           path: |
-            ${{ github.workspace }}/BinaryCache/installer/Release/arm64/rtl.arm64.msm
+            ${{ github.workspace }}/BinaryCache/installer/Release/${{ inputs.build_arch }}/rtl.arm64.msm
 
       - uses: actions/upload-artifact@v4
         with:
           name: Windows-x86-rtl-msm
           path: |
-            ${{ github.workspace }}/BinaryCache/installer/Release/x86/rtl.x86.msm
+            ${{ github.workspace }}/BinaryCache/installer/Release/${{ inputs.build_arch }}/rtl.x86.msm
```
compnerd, line 4086: This seems wrong. The output is on the associated architecture subdirectory, not the build architecture directory.

Steelskin: You're right, my bad, reverted.

.github/workflows/swift-toolchain.yml
Steelskin: If I'm holding things correctly, `ProductArchitecture` sets the output directory under `Release`. I noticed the package Tools job was not actually uploading anything because it was looking in the wrong folder. This should be set to the same value we give to `ProductArchitecture`, which is `matrix.arch` for this job.

Let's wait until the try job is done running, we'll know for sure.

Steelskin:
New try job: https://github.com/thebrowsercompany/swift-build/actions/runs/14578282345

compnerd (approved):
The smoke test failed, but the packaging succeeded. I think that this is a good step forward.

PR #968, opened 2025-05-30
Upload the VS Installer log

Steelskin:
Rather than dumping the log to the bot output, the log file is now uploaded as an artifact named after the job name.


compnerd:
`${{ github.job }}` should get you the name of the job itself. However, I think that it is better if we just limit to uploading on failure.

PR #971, opened 2025-06-02
build: Use the setup-build action in swift-toolchain

Steelskin:
* Use the `setup-build` action in every job.
* Move the build machine setup to the beginning of the job as this saves time when it fails.
* Remove `WORKAROUND_WINDOWS_SDK_VERSION` for the smoke tests, they should test the latest Windows SDK version.

### DIFF
--- a/.github/workflows/swift-toolchain.yml
+++ b/.github/workflows/swift-toolchain.yml
@@ -284,13 +284,11 @@ env:
 
   # Workaround for issues with building with SDK version 26100.
   # See https://github.com/compnerd/swift-build/issues/909 for details.
-  WORKAROUND_WINDOWS_SDK_VERSION: 10.0.22621.0
-  UNSUPPORTED_WINDOWS_SDK_VERSION: 10.0.26100.0
+  WORKAROUND_BOOTSTRAP_WINDOWS_SDK_VERSION: 10.0.22621.0
 
   # Workaround for issues with building with MSVC 14.43.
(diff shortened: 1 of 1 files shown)


Steelskin:
I ran this build with these changes last week: https://github.com/thebrowsercompany/swift-build/actions/runs/15357521575
You can ignore the failure, it fails to set the release as expected as this was a manually triggered build.

.github/workflows/swift-toolchain.yml
```diff
@@ -309,22 +307,13 @@ jobs:
...
         with:
-          path: ${{ github.workspace }}/SourceCache/swift-build
-          show-progress: false
-
-      - uses: actions/checkout@v4.2.2
-        with:
-          repository: swiftlang/swift-toolchain-sqlite
-          ref: ${{ inputs.swift_toolchain_sqlite_revision }}
-          path: ${{ github.workspace }}/SourceCache/swift-toolchain-sqlite
+          path: ${{ github.workspace }}/SourceCache/gha-swift-build
```
compnerd: How about calling this `ci-build` instead of `gha-swift-build`?

Steelskin: Done.

Steelskin:
Here's another test run with the updated paths: https://github.com/thebrowsercompany/swift-build/actions/runs/15538885807

compnerd (approved):

PR #972, opened 2025-06-04
GHA: emit `TESTING_VERSION` into `Info.plist`

compnerd:
We did not record the swift-testing version into the `Info.plist` which is expected by tools. Add this missing key to silence the warnings.


compnerd:
CC: @mhegazy

PR #973, opened 2025-06-13
Add missing attestations and input

Steelskin:
These should fix the downstream merge.

### DIFF
--- a/.github/workflows/build-toolchain.yml
+++ b/.github/workflows/build-toolchain.yml
@@ -789,6 +789,7 @@ jobs:
     permissions:
       contents: read
       id-token: write
+      attestations: write
 
     uses: ./.github/workflows/swift-toolchain.yml
     with:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-toolchain.yml
```diff
@@ -933,6 +935,7 @@ jobs:
       ANDROID_NDK_VERSION: ${{ needs.context.outputs.ANDROID_NDK_VERSION }}
       CMAKE_Swift_FLAGS: ${{ needs.context.outputs.DARWIN_CMAKE_Swift_FLAGS }}
       debug_info: ${{ fromJSON(needs.context.outputs.debug_info) }}
+      release: false
```
compnerd, line 938: Why is this unconditionally false?

Steelskin: This is for the Mac build, which is still very experimental. There is almost no risk to setting this the same way we do for Windows since the relevant jobs are disabled for macOS but I would rather err on the side of caution.

PR #974, opened 2025-06-18
setup-build: Attempt MSVC installation twice

Steelskin:
The MSVC installation sometimes fails seemingly due to the installer performing some self-update. This works around the issue by running the VS installer step twice.

A Microsoft Devcom [issue](https://developercommunity.visualstudio.com/t/Visual-Studio-Installer-randomly-fails-t/10924708) was filed for this.


compnerd (approved):
Please link in the DevComm issue before merging.

Steelskin:
This unfortunately does not work, but the DevComm issue suggested another fix, running the installer with `--noUpdateInstaller`, which I am trying right now.

PR #976, opened 2025-06-23
setup-build: Disable `vs_installer.exe` self-update

Steelskin:
As suggested on the [DevComm](https://developercommunity.visualstudio.com/t/Visual-Studio-Installer-randomly-fails-t/10924708?) thread, this adds `--noUpdateInstaller` to the `vs_installer.exe` invocation to disable the installer self-update.

This solves the issue of `vs_installer.exe` silently failing to install the requested packages.

### DIFF
--- a/.github/actions/setup-build/action.yml
+++ b/.github/actions/setup-build/action.yml
@@ -135,6 +135,7 @@ runs:
           $process = Start-Process "$VSInstaller" `
               -PassThru `
               -ArgumentList "modify", `
+                  "--noUpdateInstaller", `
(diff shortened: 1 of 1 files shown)


mhegazy (approved):

compnerd (approved):

compnerd:
Should we do this on `gha-setup-vsdenv` as well?

Steelskin:
> Should we do this on `gha-setup-vsdenv` as well?

`gha-setup-vsdenv` does not actually install anything. It just picks a VS installation with the provided packages available in case there are multiple installations.

PR #983, opened 2025-06-30
[GHA] Support running on hosts without pre-configured python 3.9

mhegazy:
Add a new GH Action to setup python using `uv`. Also adding explicit calls to setup uv where we are executing python scripts.

Downstream CI run with an image not including python (https://github.com/thebrowsercompany/swift-build/pull/302):
https://github.com/thebrowsercompany/swift-build/actions/runs/15982191821


compnerd (changes requested):
I would prefer avoiding `uv`. This is not used in the Swift.org builds, and diverging the builds further makes it harder to maintain. We already try to synchronise the build between build.ps1 and GHA, adding a diff on top is more complicated. Why can we not just fetch python if we want to use our own copy?

mhegazy:
UV is doing that for us. we are still using `python` to run commands. what this action does is download a specific python version (specified in `pyproject.toml`), and sets it in the path. we could switch to a custom action, but it will do more or less the same, find architecture, find the right package to download, and then install/expand and set in the path.

compnerd:
Yes, I realize what `uv` is being used for; I think that using `actions/setup-python` is better for that purpose. This introduces a third party dependency (and with the desire to eventually have this repository be merged into swiftlang, that complicates things further). I would rather use GitHub actions whenever there is an alternative.

mhegazy:
@compnerd reverted back to github's `actions/setup-python`

mhegazy:
@compnerd this should be updated now. downstream change passed CI on cirun and githhub: https://github.com/thebrowsercompany/swift-build/pull/302

.github/workflows/swift-toolchain.yml
```diff
@@ -681,6 +681,10 @@ jobs:
           path: ${{ github.workspace }}/SourceCache/swift
           show-progress: false
 
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
```
compnerd: Do we still build against python 3.9? I don't want to diverge on this from upstream.

mhegazy: OK. updated.  all should be 3.9 now.

mhegazy:
@compnerd one more iteration. using 3.9 now. downstream PR runs are passing at https://github.com/thebrowsercompany/swift-build/pull/302

compnerd:
Discussed offline, the global value conversion can be done as a follow up.

PR #992, opened 2025-07-18
setup-build: Update VS Installer

Steelskin:
Due to a bug in the VS installer self-update process, it sometimes fails to properly self-update, preventing the installation of the MSVC packages that we need.
In order to work around this issue, we download the newest VS installer for the installed product and run its update process before installing the MSVC packages.


compnerd (approved):
💙

PR #994, opened 2025-07-21
static-sdk: Add Distributed and Observation builds

Steelskin:


Steelskin:
Here is the internal build: https://github.com/thebrowsercompany/swift-build/actions/runs/16427788307

PR #999, opened 2025-07-31
actions: Add a build-cmake-project action

Steelskin:
This brings in a modified version of `build.ps1` from the swift repository as a custom action. This reduces the differences in the build setup between the swift repository and this repository, making it easier to keep both builds in sync.


.github/actions/build-cmake-project/action.yml
```diff
@@ -0,0 +1,117 @@
...
+  build-os:
+    description: Build operating system (e.g., "Windows", "Darwin").
+    required: true
+  build-arch:
+    description: Build architecture (e.g., "x86_64", "arm64").
+    required: true
+  os:
+    description: Operating system to build for (e.g., "Windows", "Android").
+    required: true
+  arch:
```
compnerd: Can we name these `host-os` and `host-arch`?

.github/actions/build-cmake-project/action.yml
```diff
@@ -0,0 +1,117 @@
...
+    required: false
+    default: '@{}'
+
+runs:
+  using: 'composite'
+  steps:
+    - name: Configure ${{ inputs.project-name }}
+      shell: pwsh
+      run: |
+        ${{ github.workspace }}/SourceCache/ci-build/.github/actions/build-cmake-project/Build-CMakeProject.ps1 `
```
compnerd: Could we perhaps just implement the logic here inline rather than have a separate script? We can use the same powershell content here as a single  That would allow us to then still have separate steps for the build and configure and install, which is important as that is what enables the tracing in the GHA builds.

Steelskin: I can move the file inside the script. For now, it's easier to iterate over it in a separate file, especially since I can run it locally and check the output, even if it does not run. Do keep in mind that separate steps in a re-usable action will still appear as a single "action" in the GitHub Action workflow page.

.github/actions/build-cmake-project/action.yml
Steelskin: I do not think they should, as they can be either `build`, `host` or `target` depending on the job.

.github/actions/build-cmake-project/action.yml
Steelskin: I have decided to keep this as a separate file because it makes it much easier to run the script locally to check the actual command invocation that results. Let me know if you would prefer I move it down to the yaml file and I will do that. I understand it adds a layer of indirection, but I made a point of keeping the argument order in sync so it would be easier to parse.

I have also reduced the new action to only configure the CMake project. This is to preserve separate build and install sections in the GHA UI. Unfortunately, all of the individual steps in a re-usable action always end up under a single banner, so even bringing the script down to the yaml file would not fix that.

compnerd: My issue with this approach is that it negates the benefits of the debug tracing in GHA. I would rather that we inline the script. Extracting the script to run it locally would still be relatively easy as it is just how the file is stored.

Steelskin: The script is now in the yaml file.

compnerd (approved):

PR #1003, opened 2025-08-10
Use a variable to control the version of python used in ci actions

mhegazy:
Follow up on to https://github.com/compnerd/swift-build/pull/983. This changes adds a new var to control the version of Python used in the execution of CI actions (not the version used to build lldb).

This also fixes an break introduced in the same PR on ARM64 machines by forcing the use of `x64` emulated version. 

Downstream runs:
AMD: https://github.com/thebrowsercompany/swift-build/actions/runs/16863879408
ARM: https://github.com/thebrowsercompany/swift-build/actions/runs/16863881987

### DIFF
--- a/.github/workflows/build-toolchain.yml
+++ b/.github/workflows/build-toolchain.yml
@@ -381,11 +381,7 @@ jobs:
           echo ANDROID_CLANG_VERSION=18 >> ${GITHUB_OUTPUT}
 
           # Set Python version used in the build
-          if [[ "${{ inputs.windows_build_arch }}" == "arm64" ]] ; then
-            echo python_version=3.11 >> ${GITHUB_OUTPUT}
-          else
-            echo python_version=3.9 >> ${GITHUB_OUTPUT}
-          fi
+          echo python_version=3.9.10>> ${GITHUB_OUTPUT}
 
       - uses: actions/upload-artifact@v4
         if: inputs.create_snapshot == true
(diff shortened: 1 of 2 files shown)


.github/workflows/build-toolchain.yml
```diff
@@ -379,6 +380,9 @@ jobs:
           echo ANDROID_NDK_VERSION=r27c >> ${GITHUB_OUTPUT}
           echo ANDROID_CLANG_VERSION=18 >> ${GITHUB_OUTPUT}
 
+          # Set Python version used in the build
+          echo python_version=3.9.10>> ${GITHUB_OUTPUT}
```
compnerd:
```suggestion
          echo python_version=3.9.10 >> ${GITHUB_OUTPUT}
```

.github/workflows/swift-toolchain.yml
```diff
@@ -683,7 +687,8 @@ jobs:
 
       - uses: actions/setup-python@v5
         with:
-          python-version: '3.9'
+          python-version: ${{ inputs.python_version }}
+          architecture: x64
```
compnerd: Why the hardcoded architecture?

mhegazy: ARM 64 does not have a published version of 3.9. so we are getting the emulated version. so the action does nothing. adding the arch forces it to install the x64 version, and emualtation allows it to run

compnerd: I think that we should comment that with a `FIXME` then (lets also file an issue and reference that as `FIXME(#1234) ...`). I think that as we migrate to 3.10.1 (which does have an ARM64 build as that is part of the reason for the upgrade), we should see if we can get away from the x64 variant.

mhegazy: Will do. but i think we have the same issue for 3.10. i think the first official ARM64 release was 3.11

mhegazy: `FIXME` comments added.

PR #1005, opened 2025-08-14
build: Use the configure action in the ds2 jobs

Steelskin:


.github/workflows/swift-toolchain.yml
```diff
@@ -493,6 +497,27 @@ jobs:
           setup-vs-dev-env: ${{ matrix.os == 'Windows' }}
           host-arch: ${{ matrix.arch }}
 
+      - name: Compute workspace hash
+        if: inputs.build_android
+        id: workspace_hash
+        run: |
```
compnerd, line 503: What do we need to move away from this workaround? I believe that this is needed for the differences between the self-hosted and GH hosted runners right?

Steelskin: Yes, the paths are different and that causes issues with the install targets. I have not diagnosed the problem. We could work around it by moving the hash computation into the action, which would at least make this less verbose.

PR #1009, opened 2025-08-15
build: Use the configure action in the libxml2 job

Steelskin:


Steelskin:
Downstream try job: https://github.com/thebrowsercompany/swift-build/actions/runs/16978551591

PR #1014, opened 2025-08-25
Update WindowsQuickStart.md

compnerd:
Update the commands and instructions to account for Python version change.

### DIFF
--- a/docs/WindowsQuickStart.md
+++ b/docs/WindowsQuickStart.md
@@ -36,7 +36,7 @@ del /q vs_community.exe
 
 ### Install Python
 
-The `repo` tool uses python, and as such, we need a python installation on the host. Download and install the appropriate python for your platform from https://www.python.org.
+The `repo` tool uses Python, and as such, we need a Python installation on the host. We recommend installing **Python 3.9.10** to ensure compatibility with the provided scripts and examples. Download and install Python 3.9.10 for your platform from [https://www.python.org/downloads/release/python-3910/](https://www.python.org/downloads/release/python-3910/).
 
 ### Enable Symbolic Links Support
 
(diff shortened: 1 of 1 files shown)


docs/WindowsQuickStart.md
```diff
@@ -192,7 +192,7 @@ The following content in your Powershell profile file (whose path is stored in t
 function Set-SwiftEnv {
   $SwiftRoot = "S:\Program Files\Swift"
   $env:SDKROOT = "${SwiftRoot}\Platforms\Windows.platform\Developer\SDKs\Windows.sdk"
-  $env:Path = "${env:ProgramFiles}\Python39;${SwiftRoot}\Runtimes\0.0.0\usr\bin;${SwiftRoot}\Toolchains\0.0.0+Asserts\usr\bin;${env:Path}"
+  $env:Path = "S:\b\Python${env:PROCESSOR_ARCHITECTURE}-3.9.10\tools;${SwiftRoot}\Runtimes\0.0.0\usr\bin;${SwiftRoot}\Toolchains\0.0.0+Asserts\usr\bin;${env:Path}"
```
mhegazy, line 195: i am guessing this is not intentional

compnerd: That is currently intentional ... it seems that my local build is still using 3.9.10? I think that we need to figure out how to figure out how to get the right Python library into the path.

mhegazy (approved):

docs/WindowsQuickStart.md
mhegazy: but build.ps1 should be extracting 3.10.1 and not 3.9.

compnerd: Yeah, it should be. We can update it once I have that working locally (or can point to the merged PR).

PR #1023, opened 2025-09-08
Build online installer

mhegazy:
This changes builds the online flavor of the installer and uploads the binaries to a public R2 storage bucket. The offline installer is still available. 

## Changes: 
- Build both online and offline installer
- upload cabs and msi files to R2 storage and use the url to build the installer
- Use the online installer for smoke testing
- Change name of payloads accordingly

## Testing
- testing the resulting installer localy
- smoke testing with the new installer

**Successful downstream job (https://github.com/thebrowsercompany/swift-build/actions/runs/17561511426) ** (Note: i still need to test changes to upload files on the release)

### DIFF
--- a/.github/workflows/build-toolchain.yml
+++ b/.github/workflows/build-toolchain.yml
@@ -145,7 +145,12 @@ on:
         required: true
       PASSPHRASE:
         required: true
-
+      R2_ACCOUNT_ID:
+        required: true
+      R2_ACCESS_KEY_ID:
+        required: true
+      R2_SECRET_ACCESS_KEY:
+        required: true
 
 jobs:
   context:
@@ -871,6 +876,9 @@ jobs:
       SYMBOL_SERVER_PAT: ${{ secrets.SYMBOL_SERVER_PAT }}
       CERTIFICATE: ${{ secrets.CERTIFICATE }}
       PASSPHRASE: ${{ secrets.PASSPHRASE }}
+      R2_ACCOUNT_ID: ${{ secrets.R2_ACCOUNT_ID }}
+      R2_ACCESS_KEY_ID: ${{ secrets.R2_ACCESS_KEY_ID }}
+      R2_SECRET_ACCESS_KEY: ${{ secrets.R2_SECRET_ACCESS_KEY }}
 
   mac-build:
     # TODO: Enable the mac build.
@@ -955,6 +963,9 @@ jobs:
       SYMBOL_SERVER_PAT: ${{ secrets.SYMBOL_SERVER_PAT }}
       CERTIFICATE: ${{ secrets.CERTIFICATE }}
       PASSPHRASE: ${{ secrets.PASSPHRASE }}
+      R2_ACCOUNT_ID: ${{ secrets.R2_ACCOUNT_ID }}
+      R2_ACCESS_KEY_ID: ${{ secrets.R2_ACCESS_KEY_ID }}
+      R2_SECRET_ACCESS_KEY: ${{ secrets.R2_SECRET_ACCESS_KEY }}
 
   snapshot:
     runs-on: ubuntu-latest
@@ -998,6 +1009,7 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
+          # TODO: Upload the offline installer as well.
           branch_version_string=${{ inputs.swift_version || '0.0.0' }}
           if [[ $branch_version_string == "0.0.0" ]]; then
             latest="true"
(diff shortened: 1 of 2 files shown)


.github/workflows/build-toolchain.yml
```diff
@@ -986,41 +997,51 @@ jobs:
...
-          # ARM64
-          cd ${{ github.workspace }}/tmp/arm64
-
-          mv installer.exe installer-arm64.exe
-          gh release upload ${{ needs.context.outputs.swift_tag }} installer-arm64.exe -R ${{ github.repository }}
-
-          shasum -a 256 installer-arm64.exe > installer-arm64.exe.sha256
-          gh release upload ${{ needs.context.outputs.swift_tag }} installer-arm64.exe.sha256 -R ${{ github.repository }}
+          # Upload all installer variants
+          upload_installer "${{ github.workspace }}/tmp/online/amd64/installer.exe" "installer-amd64.exe"
```
compnerd, line 1044: Why name these to `installer-amd64` and the existing ones to `installer-offline-amd64`?

.github/workflows/swift-toolchain.yml
```diff
@@ -5091,15 +5097,92 @@ jobs:
...
+          subject-path: ${{ github.workspace }}/BinaryCache/installer/Release/${{ matrix.arch }}/installer.exe
+
+      - name: Upload installer (online)
+        if: ${{ inputs.release }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: Windows-${{ matrix.arch }}-installer-online
+          path: ${{ github.workspace }}/BinaryCache/installer/Release/${{ matrix.arch }}/installer.exe
+
+      - name: Upload msi's and cab's
```
compnerd, line 5176:
```suggestion
      - name: Upload MSIs and CABs 
```

.github/workflows/swift-toolchain.yml
```diff
@@ -5091,15 +5097,92 @@ jobs:
...
+      - name: Upload installer (online)
+        if: ${{ inputs.release }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: Windows-${{ matrix.arch }}-installer-online
+          path: ${{ github.workspace }}/BinaryCache/installer/Release/${{ matrix.arch }}/installer.exe
+
+      - name: Upload msi's and cab's 
+        if: ${{ inputs.release }}
+        uses: ryand56/r2-upload-action@b801a390acbdeb034c5e684ff5e1361c06639e7c # v1.4
```
compnerd: Is this absolutely required? Is there an official action from cloudflare or GH?

.github/workflows/build-toolchain.yml
mhegazy: to make this one the default. tools that install from the release look for `installer-amd64`. do you want to keep the offline version as the default?

.github/workflows/swift-toolchain.yml
mhegazy: applied

.github/workflows/swift-toolchain.yml
mhegazy: there is no gh or cloudeflare action either. but i think we can use AWS s3 action if we want. i beleive R2 is api comatible with S3, that is what i used localy.

the other option is to use the API directly, not a lot of bells and wisles but will do the job.

Steelskin (approved):
This looks good to me, modulo the open question on which action to use.

.github/workflows/swift-toolchain.yml
compnerd: I think that the AWS action might be easier to convince others to use.

.github/workflows/build-toolchain.yml
compnerd: Yeah, lets do that - lets keep the offline version as the old name, although I would be open to renaming everything to have uniform naming.

.github/workflows/swift-toolchain.yml
mhegazy: moved to S3 upload. need to do a full test, but need to unbreak the build first :)

.github/workflows/swift-toolchain.yml
```diff
@@ -5091,26 +5097,109 @@ jobs:
...
+        with:
+          name: Windows-${{ matrix.arch }}-installer-offline
+          path: ${{ github.workspace }}/BinaryCache/installer/Release/${{ matrix.arch }}/installer.exe
+    
+      - name: Generate online installer download URL
+        if: ${{ inputs.release }}
+        id: generate-base-release-download-prefix
+        run: |
+          $BaseReleaseDownloadUrlPrefix = "${{ inputs.swift_tag }}/${{ matrix.arch }}/${{ github.run_id }}"
+          echo "url_prefix=$BaseReleaseDownloadUrlPrefix" >> $env:GITHUB_OUTPUT
```
tristanlabelle: Minor: without https:// it's not really a prefix but rather a slug (url subcomponent).

tristanlabelle (approved):

PR #1031, opened 2025-09-16
Add smoke tests to ensure redistributables have expected DLLs

speednoisemovement:
This change
- Saves the names of all .dlls in the runtime's `/usr/bin` for all architectures
- For all experimental SDK rtl MSMs:
  - If shared, checks that all .dlls from the runtime are present
  - If static, checks that all .dlls on a hardcoded list are present

### DIFF
--- a/.github/workflows/swift-toolchain.yml
+++ b/.github/workflows/swift-toolchain.yml
@@ -4354,6 +4354,10 @@ jobs:
     name: Package Windows SDK & Runtime
     needs: [stdlib, sdk, experimental-sdk]
     runs-on: ${{ inputs.default_build_runner }}
+    outputs:
+      expected-dlls-amd64: ${{ steps.write-expectations.outputs.expected-dlls-amd64}}
+      expected-dlls-arm64: ${{ steps.write-expectations.outputs.expected-dlls-arm64}}
+      expected-dlls-x86: ${{ steps.write-expectations.outputs.expected-dlls-x86}}
 
     steps:
       - uses: actions/checkout@v4.2.2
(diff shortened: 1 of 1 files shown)


.github/workflows/swift-toolchain.yml
```diff
@@ -4354,6 +4354,10 @@ jobs:
     name: Package Windows SDK & Runtime
     needs: [stdlib, sdk, experimental-sdk]
     runs-on: ${{ inputs.default_build_runner }}
+    outputs:
+      expected-dlls-amd64: ${{ steps.write-expectations.outputs.expected-dlls-amd64}}
+      expected-dlls-arm64: ${{ steps.write-expectations.outputs.expected-dlls-arm64}}
+      expected-dlls-x86: ${{ steps.write-expectations.outputs.expected-dlls-x86}}
```
compnerd:
```suggestion
      expected-dlls-amd64: ${{ steps.write-expectations.outputs.expected-dlls-amd64 }}
      expected-dlls-arm64: ${{ steps.write-expectations.outputs.expected-dlls-arm64 }}
      expected-dlls-x86: ${{ steps.write-expectations.outputs.expected-dlls-x86 }}
```

.github/workflows/swift-toolchain.yml
```diff
@@ -4626,6 +4630,20 @@ jobs:
               -p:WindowsExperimentalRuntimeX64="${{ github.workspace }}/BuildRoot/Library/Developer/Runtimes.Experimental/Windows-x86_64" `
               -p:WindowsExperimentalRuntimeX86="${{ github.workspace }}/BuildRoot/Library/Developer/Runtimes.Experimental/Windows-i686" `
               ${{ github.workspace }}/SourceCache/swift-installer-scripts/platforms/Windows/platforms/windows/windows.wixproj
+      - name: Write DLL Expectations
+        id: write-expectations
+        run: |
+          $runtimes_amd64 = "${{ github.workspace }}/BuildRoot/Library/Developer/Runtimes.Experimental/Windows-x86_64/usr/bin"
+          $dlls_amd64 = Get-ChildItem -Path $runtimes_amd64 -Filter *.dll -File | ForEach-Object Name | ConvertTo-Json -Compress
```
compnerd: Why the conversion to JSON?

.github/workflows/swift-toolchain.yml
```diff
@@ -5124,6 +5142,59 @@ jobs:
...
+        ) || '["BlocksRuntime.dll", "dispatch.dll"]' }}
+    steps:
+      - name: Download MSM
+        uses: actions/download-artifact@v4
+        with:
+          name: ${{ env.MSM_ARTIFACT_NAME }}
+          path: ${{ github.workspace }}/tmp
+      - name: Test ${{ matrix.variant }} ${{ matrix.arch }}
+        run: |
+          $tmp = Join-Path $env:GITHUB_WORKSPACE 'tmp'
```
compnerd: Why not just use `$env:TEMP`?

.github/workflows/swift-toolchain.yml
speednoisemovement: Was being defensive about pulling it out of the output. Can I just dump the ForEach results in there, newlines and all?

.github/workflows/swift-toolchain.yml
speednoisemovement: AFAICT $env:TEMP is an actual temp dir. This is `${{ github.workspace }}/tmp/` following other uses of download-artifact

speednoisemovement: Would it make more sense to just download to `env.Temp` in the first place?

.github/workflows/swift-toolchain.yml
compnerd: I believe you should be able to do that.

.github/workflows/swift-toolchain.yml
compnerd: Yeah, I think that it makes sense to download to there.

speednoisemovement: Used runner_temp since the PowerShell `$env:EMP` doesn't seem like it's usable in the action argument context

.github/workflows/swift-toolchain.yml
speednoisemovement: I did some testing in a sandbox and just writing the results directly gives you a flat string. I tried splitting it by whitespace, which works fine as long as none of the filenames have spaces.

Options:
1) Keep JSON
2) Split on whitespace and rely on none of the DLLs having whitespace in the names (seems likely but not 100% guaranteed)
3) Something I'm not thinking of?

speednoisemovement: Summarizing offline conversation: keeping JSON but added a comment to explain. Also reworked this a little to reduce repetition

compnerd (approved):

PR #1033, opened 2025-09-24
Update gha-setup-swift to latest

mhegazy:
We are seeing failures installing swift (see [example](https://github.com/thebrowsercompany/swift-build/actions/runs/17976216297/job/51133301454)). newer versions of the action collect logs, so getting us to be on the cutting edge

Downstream run: https://github.com/thebrowsercompany/swift-build/actions/runs/17985190166


mhegazy:
looks like my downstream run is triggering the same issue. looking into it.

compnerd (approved):

PR #1035, opened 2025-10-01
Build Brotli in advance of it being a Foundation dep

speednoisemovement:
Upcoming change: https://github.com/swiftlang/swift/pull/83441

### DIFF
--- a/.github/workflows/swift-toolchain.yml
+++ b/.github/workflows/swift-toolchain.yml
@@ -1408,7 +1408,7 @@ jobs:
           path: ${{ github.workspace }}/BuildRoot/Library/brotli-${{ inputs.brotli_version }}/usr
 
   curl:
-    needs: [zlib, brotli]
+    needs: [zlib]
     runs-on: ${{ inputs.default_build_runner }}
 
     strategy:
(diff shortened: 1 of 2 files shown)


.github/workflows/swift-toolchain.yml
```diff
@@ -1479,6 +1579,9 @@ jobs:
               'USE_WIN32_LDAP' = "NO";
               'ZLIB_ROOT' = "${{ github.workspace }}/BuildRoot/Library/zlib-${{ inputs.zlib_version }}/usr";
               'ZLIB_LIBRARY' = "${{ github.workspace }}/BuildRoot/Library/zlib-${{ inputs.zlib_version }}/usr/lib/zlibstatic.lib";
+              'BROTLI_INCLUDE_DIR' = "${{ github.workspace }}/BuildRoot/Library/brotli-${{ inputs.brotli_version }}/usr/include";
```
speednoisemovement: The open build.ps1 PR uses `BROTLI_DIR` but I couldn't get that to work 🤷

compnerd: That relies on the CMake configuration from brotli. That should be available in the build tree, not sure if it is in the install tree.

speednoisemovement: This would be something like `BrotliConfig.cmake`? I don't see one of those in my test build.

speednoisemovement: OK, I see what's going on here: BROTLI_DIR is a red herring (and you can even see in the output on Swift.org CI that it's ignored). It works because both Brotli and cURL have the same install prefix. I think we need to do it this way in GHA where nothing shares an install dir.

speednoisemovement: Per offline conversation, removed cURL integration for now. We will try to land config changes to upstream `brotli` to allow `BROTLI_DIR` to work.

PR #1038, opened 2025-10-03
Use Azure Trusted Signing when available

speednoisemovement:
🚨 Do not merge: Depends on https://github.com/swiftlang/swift-installer-scripts/pull/463 🚨

This change uses [Azure Trusted Signing](https://learn.microsoft.com/en-us/azure/trusted-signing/) to sign all WiX-based outputs if:
1. `inputs.signed` is true
2. The `TRUSTED_SIGNING_ACCOUNT` secret is set for the repository.

The following secrets are required:
- `AZURE_SP_CREDENTIALS`: [JSON credentials](https://github.com/Azure/login?tab=readme-ov-file#login-with-a-service-principal-secret) for an Azure Service Principal with [Trusted Signing Certificate Profile Signer](https://learn.microsoft.com/en-us/azure/trusted-signing/concept-trusted-signing-resources-roles#supported-roles) roles on the given trusted signing account and certificate profiles
- `TRUSTED_SIGNING_ACCOUNT`, an [Azure Trusted Signing Account](https://learn.microsoft.com/en-us/azure/trusted-signing/quickstart?tabs=registerrp-portal%2Caccount-portal%2Ccertificateprofile-portal%2Cdeleteresources-portal#create-a-trusted-signing-account) which owns `TRUSTED_SIGNING_PROD_PROFILE` (and `TRUSTED_SIGNING_TEST_PROFILE` when used)
- `TRUSTED_SIGNING_PROD_PROFILE`, a non-test [certificate profile](https://learn.microsoft.com/en-us/azure/trusted-signing/concept-trusted-signing-resources-roles#certificate-profiles)

and optionally
- `TRUSTED_SIGNING_TEST_PROFILE`, which isn't actively used in the workflow, but is helpful to have in development.

### DIFF
--- a/.github/workflows/swift-toolchain.yml
+++ b/.github/workflows/swift-toolchain.yml
@@ -4177,7 +4177,7 @@ jobs:
           '{
             "Endpoint": "https://eus.codesigning.azure.net",
             "CodeSigningAccountName": "${{ secrets.TRUSTED_SIGNING_ACCOUNT }}",
-            "CertificateProfileName": "${{ secrets.TRUSTED_SIGNING_TEST_PROFILE }}",
+            "CertificateProfileName": "${{ secrets.TRUSTED_SIGNING_PROD_PROFILE }}",
             "ExcludeCredentials": [
               "ManagedIdentityCredential",
               "WorkloadIdentityCredential",
@@ -4795,7 +4795,7 @@ jobs:
           '{
             "Endpoint": "https://eus.codesigning.azure.net",
             "CodeSigningAccountName": "${{ secrets.TRUSTED_SIGNING_ACCOUNT }}",
-            "CertificateProfileName": "${{ secrets.TRUSTED_SIGNING_TEST_PROFILE }}",
+            "CertificateProfileName": "${{ secrets.TRUSTED_SIGNING_PROD_PROFILE }}",
             "ExcludeCredentials": [
               "ManagedIdentityCredential",
               "WorkloadIdentityCredential",
@@ -5128,7 +5128,7 @@ jobs:
           '{
             "Endpoint": "https://eus.codesigning.azure.net",
             "CodeSigningAccountName": "${{ secrets.TRUSTED_SIGNING_ACCOUNT }}",
-            "CertificateProfileName": "${{ secrets.TRUSTED_SIGNING_TEST_PROFILE }}",
+            "CertificateProfileName": "${{ secrets.TRUSTED_SIGNING_PROD_PROFILE }}",
             "ExcludeCredentials": [
               "ManagedIdentityCredential",
               "WorkloadIdentityCredential",
@@ -5310,7 +5310,7 @@ jobs:
           '{
             "Endpoint": "https://eus.codesigning.azure.net",
             "CodeSigningAccountName": "${{ secrets.TRUSTED_SIGNING_ACCOUNT }}",
-            "CertificateProfileName": "${{ secrets.TRUSTED_SIGNING_TEST_PROFILE }}",
+            "CertificateProfileName": "${{ secrets.TRUSTED_SIGNING_PROD_PROFILE }}",
             "ExcludeCredentials": [
               "ManagedIdentityCredential",
               "WorkloadIdentityCredential",
--- a/.github/workflows/swift-toolchain.yml
+++ b/.github/workflows/swift-toolchain.yml
@@ -264,23 +264,6 @@ on:
 
 env:
   PINNED_BOOTSTRAP_TOOLCHAIN_VERSION: 6.1.2
-  TRUSTED_SIGNING_METADATA_JSON: |
-    {
-      "Endpoint": "https://eus.codesigning.azure.net",
-      "CodeSigningAccountName": "${{ secrets.TRUSTED_SIGNING_ACCOUNT }}",
-      "CertificateProfileName": "${{ secrets.TRUSTED_SIGNING_TEST_PROFILE }}",
-      "ExcludeCredentials": [
-        "ManagedIdentityCredential",
-        "WorkloadIdentityCredential",
-        "SharedTokenCacheCredential",
-        "VisualStudioCredential",
-        "VisualStudioCodeCredential",
-        "EnvironmentCredential",
-        "AzurePowerShellCredential",
-        "AzureDeveloperCliCredential",
-        "InteractiveBrowserCredential"
-      ]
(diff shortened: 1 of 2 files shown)


.github/workflows/swift-toolchain.yml
```diff
@@ -3896,11 +3904,38 @@ jobs:
...
           path: ${{ github.workspace }}/BuildRoot/Library
 
+  # Determines whether we're using trusted signing in this repository. If so and we're signing, downloads the trusted signing DLL to reduce boilerplate in steps that use it.
+  # Setting an output is necessary because the `env` and `secrets` contexts are not always available in `if` clauses.
+  configure_signing:
+    if: inputs.build_os == 'Windows'
+    name: Configure signing
+    runs-on: ${{ inputs.default_build_runner }}
+    outputs:
+      uses_trusted_signing: ${{ steps.test_trusted_signing.outputs.uses_trusted_signing }}
```
Steelskin, line 3914: I don't think this is necessary. Presumably, if we have selected `signed` on the build workflow, the repository is already configured properly. Otherwise, it is a user error and we should eventually error out.

.github/workflows/swift-toolchain.yml
```diff
@@ -4116,6 +4151,47 @@ jobs:
...
+      - uses: actions/download-artifact@v4
+        if: inputs.signed && needs.configure_signing.outputs.uses_trusted_signing == 'true'
+        with:
+          name: trusted-signing-dll
+          path: ${{ runner.temp }}/trusted-signing-dll
+
+      - name: Prepare trusted signing arguments
+        if: inputs.signed && needs.configure_signing.outputs.uses_trusted_signing == 'true'
+        run: |
+          # We're unconditionally using x64 because there's no arm64 version of signtool.exe as of Sep 2025.
```
Steelskin: nit: Oct :p

.github/workflows/swift-toolchain.yml
```diff
@@ -3896,11 +3904,38 @@ jobs:
...
+
+    steps:
+      - name: Test whether we're using trusted signing
+        id: test_trusted_signing
+        run: |
+          "uses_trusted_signing=${{ secrets.TRUSTED_SIGNING_ACCOUNT != ''}}" | Out-File -FilePath ${env:GITHUB_OUTPUT} -Encoding utf8 -Append
+      - uses: nuget/setup-nuget@v2
+        if: inputs.signed && steps.test_trusted_signing.outputs.uses_trusted_signing == 'true'
+      - run: |
+          nuget install -x -OutputDirectory ${{ github.workspace }} Microsoft.Trusted.Signing.Client
```
Steelskin, line 3924: How long does this step take? Is it worth it to not re-run it on every jon?

.github/workflows/swift-toolchain.yml
```diff
@@ -4116,6 +4151,47 @@ jobs:
             & "${{ github.workspace }}\SourceCache\mimalloc\bin\minject$BuildSuffix" -l "$Exe"
           }
 
+      - uses: azure/login@v2
```
Steelskin, line 4154: It might be worth it to move all of this to a testable reusable action. It's quite a bit of duplicated code and a test would be handy to figure out what is broken when things break.

Steelskin (commented):
Thank you! So happy to see this finally happening.

.github/workflows/swift-toolchain.yml
speednoisemovement: We need this if we want to keep the explicit cert code path around, so that we can tell whether we're doing trusted or "regular".

.github/workflows/swift-toolchain.yml
speednoisemovement: Not that long, just trying to reduce duplication, as you noticed below

.github/workflows/swift-toolchain.yml
speednoisemovement: Agree, but my understanding was we have a pretty high bar for moving things out into reusable actions?

Steelskin: @compnerd How would you like to proceed? I think there is value in making this a separate re-usable sub-action:
* Easier maintenance burden when we need to update the code. It's very easy to miss a copy of the duplicated code.
* We can add a simple test for the re-usable action so when signing break, we can use that to test signing in isolation without needing to either create a new dummy workflow for testing purposes or running the whole build again.

WDYT?

.github/workflows/swift-toolchain.yml
Steelskin: I am also concerned about Microsoft changing the layout at some point so all of the necessary resource files aren't in this directory but that's probably good enough for now. `nuget` has been flaky in the past too so reducing it to a single command is not a bad idea, we can also fail early.

.github/workflows/swift-toolchain.yml
compnerd: I think it is better to avoid the custom sub-actions; in general, we should be shying away from them - it is better to make it easier to dive into the build than to make the build "maintainable". We don't want the build to be someones "full time job".

.github/workflows/swift-toolchain.yml
```diff
@@ -4116,6 +4151,47 @@ jobs:
...
+        with:
+          name: trusted-signing-dll
+          path: ${{ runner.temp }}/trusted-signing-dll
+
+      - name: Prepare trusted signing arguments
+        if: inputs.signed && needs.configure_signing.outputs.uses_trusted_signing == 'true'
+        run: |
+          # We're unconditionally using x64 because there's no arm64 version of the DLL as of Oct 2025.
+          # TODO: Update with more info once we've filed a bug against Microsoft.
+          $signtoolPath = Join-Path -Path ${env:WindowsSdkVerBinPath} -ChildPath "x64/"
```
compnerd, line 4170: Does this work under PRISM? That is, do we need to limit this step to if the windows host is x64?

compnerd (approved):
If you want to reduce duplication, one option would be to create the signing metadata once and share that across the jobs in the setup for the signing.

.github/workflows/swift-toolchain.yml
speednoisemovement: Yes, that's how I had to do it locally :(

speednoisemovement: As in yes it works, not yes we have to limit it

PR #1058, opened 2025-11-30
Extract common signing steps to an action

mhegazy:
Signing steps are duplicated between 4 jobs now. extract them to an action to avoid duplication.

downstream job without signing: https://github.com/thebrowsercompany/swift-build/actions/runs/19777383994
downstream jobs with signing: https://github.com/thebrowsercompany/swift-build/actions/runs/19774595404
downstream 6.3 job: https://github.com/thebrowsercompany/swift-build/actions/runs/19804026228/job/56735897976


sundresh (approved):
I like the deduplication here.

Looking at this PR, I also see a lot of duplication in the `msbuild` steps. It might be worth factoring that out later (not asking for it in this PR), although I'm not sure if the calls would end up just as long in order to pass in all of the arguments from the calling context.

PR #1103, opened 2026-04-13
Smoke tests: Use native build system

Steelskin:
`swift build` now uses the xcode build system internally by default. The swift-build build system currently lacks support for custom SDKs, breaking our tests. Use `--build-system native` to work around this until the underlying issue is fixed.


compnerd (approved):
The change is fine as a workaround, but it isn't "our usage" that is the issue, the problem is that the swift-build build system is not fully well tested yet, and we are catching issues in swift-build. Please update the commit message before merging.

PR #1104, opened 2026-04-14
Smoke tests: Use the native build system for tests

Steelskin:
The swift-build build system builds test binaries as `.exe` files, rather than as `.xctest` files. This breaks some test expectations in `swift-issue-reporting`. Build and run the tests with the native build system to work around the issue.

See swiftlang/swift-package-manager#9940 for details.


compnerd (approved):
Please file and tag an issue

Steelskin:
I filed swiftlang/swift-package-manager#9940 for this.

PR #1118, opened 2026-05-18
CAS: delete archive when restoring

speednoisemovement:
We're having a disk space issue right now, and I figured this was an inoffensive way to free up some space while we investigate.

Also no longer prints the full file listing of the CAS dir to clean up the logs.


hjyamauchi (approved):
I was in parallel testing something similar https://github.com/compnerd/swift-build/pull/1119

You're welcome to pick whichever

PR #1122, opened 2026-06-22
Remove Mac build

speednoisemovement:
We have no short/medium term plans to make this work, and having it dormant in the workflows is noisy.  It's almost certainly rotted, so when we want to do this for real, the current setup will need to be redone anyway.

### DIFF
--- a/.github/actions/configure-cmake-project/action.yml
+++ b/.github/actions/configure-cmake-project/action.yml
@@ -40,7 +40,7 @@ inputs:
     required: false
     default: false
   build-os:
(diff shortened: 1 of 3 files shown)


compnerd (approved):
I wonder if we should also further simplify and remove `build-toolchain.yml` and fold back down to just `swift-toolchain.yml`.

speednoisemovement:
> I wonder if we should also further simplify and remove `build-toolchain.yml` and fold back down to just `swift-toolchain.yml`.

Sounds like a good idea. Let me look into details.

PR #1124, opened 2026-07-28
Bootstrapping: Update configure-cmake-project

Steelskin:
This brings in the latest changes from upstream build.ps1 into the configure-cmake-project action. This is a needed step to bring in the bootstrapping changes into GHA. While the call sites have been updated to match the new parameters for the action, it is a known expectation that these changes will break the build. Though, due to the bootstrapping changes in upstream Swift, the build is already broken. In particular, the changes will now reference non-existent compilers, until the bootstrapping changes are brought into GHA.

* Bring in all of the latest changes from upstream build.ps1.
* Use CMake 4.4 everywhere.
* Split the action PowerShell into a separate file as there is a character limit in the GHA yaml files.
* Update call sites to roughly match the new format.

### DIFF
--- a/.github/actions/configure-cmake-project/action.ps1
+++ b/.github/actions/configure-cmake-project/action.ps1
@@ -597,7 +597,7 @@ switch ($OS) {
     }
 }
 
-# Note: On GHA, sccache is no longer supported.
+# Note: In build.ps1, sccache is no longer supported.
 if ($EnableCaching -and $OS -ne "Android") {
     if ($UseC) {
         Add-KeyValueIfNew $Defines CMAKE_C_COMPILER_LAUNCHER "sccache"
--- a/.github/actions/configure-cmake-project/action.ps1
+++ b/.github/actions/configure-cmake-project/action.ps1
@@ -1,0 +1,698 @@
+# Configure a CMake project, mimicking `Build-CMakeProject` from the Swift
+# repository's `utils\build.ps1`. See `action.yml` for the documented list of
+# differences between this action and `build.ps1`.
+#
+# This script is invoked by `action.yml`. All inputs are passed via `INPUT_*`
+# environment variables (see `action.yml`), so that no GitHub Actions
+# `${{ }}` template syntax leaks into this file - GHA does not template `.ps1`
+# files, only the inline `run:` block in YAML.
+
+Remove-Item env:\SDKROOT -ErrorAction SilentlyContinue
+$ExeSuffix = if ($IsWindows) { ".exe" } else { "" }
+
+function ConvertTo-Bool([string]$Value) {
+    if ([string]::IsNullOrWhiteSpace($Value)) { return $false }
+    return [System.Convert]::ToBoolean($Value)
+}
+
+$ProjectName = $env:INPUT_PROJECT_NAME
+$SwiftVersion = $env:INPUT_SWIFT_VERSION
+$EnableCaching = ConvertTo-Bool $env:INPUT_ENABLE_CACHING
+$EnableCAS = ConvertTo-Bool $env:INPUT_ENABLE_CAS
+$DebugInfo = ConvertTo-Bool $env:INPUT_DEBUG_INFO
+$BuildOS = $env:INPUT_BUILD_OS
+$BuildArch = $env:INPUT_BUILD_ARCH
+$OS = $env:INPUT_OS
+$Arch = $env:INPUT_ARCH
+$SrcDir = $env:INPUT_SRC_DIR
+$BinDir = $env:INPUT_BIN_DIR
+$InstallDir = $env:INPUT_INSTALL_DIR
+$AndroidAPILevel = $env:INPUT_ANDROID_API_LEVEL
+$AndroidClangVersion = $env:INPUT_ANDROID_CLANG_VERSION
+$NDKPath = $env:INPUT_NDK_PATH
+$SwiftSDK = $env:INPUT_SWIFT_SDK_PATH
+$CacheScript = $env:INPUT_CACHE_SCRIPT
+$UseHostToolchain = ConvertTo-Bool $env:INPUT_USE_HOST_TOOLCHAIN
+$UseASM_MASM = ConvertTo-Bool $env:INPUT_USE_ASM_MASM
+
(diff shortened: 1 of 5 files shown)


.github/actions/configure-cmake-project/action.yml
```diff
@@ -31,12 +29,16 @@ inputs:
...
+    description: Enable cas.
+    required: false
+    default: false
   debug-info:
     description: Enable debug information in the build.
     required: false
     default: false
   build-os:
-    description: Build operating system (e.g., "Windows").
+    description: Build operating system (e.g., "Windows", "Darwin").
```
speednoisemovement: Revert this, we dropped the macOS build

.github/actions/configure-cmake-project/action.yml
```diff
@@ -31,12 +29,16 @@ inputs:
     description: Enable sccache.
     required: false
     default: false
+  enable-cas:
```
speednoisemovement: Revert, we can add it back when we readd CAS. It's confusing IMO since someone looking can't tell if CAS is or isn't supported

Or alternatively, add a TODO comment or something, so it's clear this is inert for now?

.github/actions/configure-cmake-project/action.yml
```diff
@@ -73,31 +75,48 @@ inputs:
...
-  use-gnu-driver:
-    description: Use the GNU driver for building the project.
+    default: ''
+  swift-compiler:
+    description: |
+      Swift compiler selector, resolved against the script's `$Compilers` table
+      (e.g. "Pinned.Swift", "Stage1.Swift"). Empty means no Swift compiler.
+    required: false
+    default: ''
+  use-host-toolchain:
```
speednoisemovement: Maybe not for this PR: should we consider renaming this given the description below?

.github/actions/configure-cmake-project/action.ps1
```diff
@@ -0,0 +1,698 @@
...
+            AssumeFunctional = $true
+        }
+    }
+}
+
+# Mirror `build.ps1`'s `$Compilers.Host` trick: a computed pseudo-set that
+# redirects to the host MSVC toolchain or the pinned toolchain depending on
+# `$UseHostToolchain`. This keeps the host-vs-pinned decision in one place so
+# call sites can simply select `Host.C` / `Host.CXX`. Note there is no
+# `Host.Swift`: Swift always comes from a real stage.
```
speednoisemovement, line 240: Could we somehow enforce this in `Resolve-Tool` just as a fail-fast in case something accidentally tries to do this?

.github/actions/configure-cmake-project/action.yml
```diff
@@ -31,12 +29,16 @@ inputs:
     description: Enable sccache.
```
speednoisemovement, line 29: I think we're probably safe to kill this whole sccache bit

.github/actions/configure-cmake-project/action.ps1
```diff
@@ -0,0 +1,698 @@
...
+        Add-FlagsDefine $Defines CMAKE_Swift_FLAGS @(
+            "-explicit-module-build",
+            "-cache-compile-job",
+            "-cas-path", $CasPath,
+            "-incremental-dependency-scan"
+        )
+    }
+}
+
+# Note: On GHA, sccache is no longer supported.
```
speednoisemovement: So why keep it?

speednoisemovement (approved):
LGTM!

.github/actions/configure-cmake-project/action.yml
Steelskin: I thought we had re-enabled it? I'd rather do that in a follow-up PR, though.

.github/actions/configure-cmake-project/action.yml
Steelskin: I renamed to "use-msvc-host-toolchain"

.github/actions/configure-cmake-project/action.ps1
Steelskin: Shouldn't that already be the case?
In the second iteration of the loop `$Part` should resolve to `Swift`. At that point `$Node.Contains($Part)` should be false, so we'll throw.

.github/actions/configure-cmake-project/action.ps1
Steelskin: The comment was incorrect, sorry!

.github/actions/configure-cmake-project/action.yml
speednoisemovement: We haven't yet. I was thinking about it but want to try the Azure Blob Storage CAS approach first. Anyway, I was overindexing on the incorrect " Note: On GHA, sccache is no longer supported" comment, so withdrawn!

compnerd (commented):
I'm not sure how I feel about this - this is **A LOT** of additional cognitive burden for IMO low gain. The externalisation means that now you need to track through the yaml and the powershell and the CMake and ninja. This is adding more indirection that I'm generally not a fan of.

Steelskin:
> I'm not sure how I feel about this - this is **A LOT** of additional cognitive burden for IMO low gain. The externalisation means that now you need to track through the yaml and the powershell and the CMake and ninja. This is adding more indirection that I'm generally not a fan of.

We have the problem that there is a hard limit on the number of characters for a step inner code, which is what this separation aims to work around. That being said, we could cut the middle man, remove the action and call the PowerShell script directly. WDYT?

compnerd:
> > I'm not sure how I feel about this - this is **A LOT** of additional cognitive burden for IMO low gain. The externalisation means that now you need to track through the yaml and the powershell and the CMake and ninja. This is adding more indirection that I'm generally not a fan of.
> 
> 
> 
> We have the problem that there is a hard limit on the number of characters for a step inner code, which is what this separation aims to work around. That being said, we could cut the middle man, remove the action and call the PowerShell script directly. WDYT?

I'm not a fan of that either. The goal is to get to a declarative build. The call out to a script "works" in the sense that you can ride the train or have a Ferrari for transportation. The loss of functionality from the scripts I'm not a fan of.

PR #1125, opened 2026-08-04
Introduce Azure Key Vault-based signing

speednoisemovement:
Also removes trusted signing which no longer works (which makes this net negative LOC!)

### DIFF
--- a/.github/workflows/swift-toolchain.yml
+++ b/.github/workflows/swift-toolchain.yml
@@ -317,7 +317,7 @@ on:
 
 env:
   PINNED_BOOTSTRAP_TOOLCHAIN_VERSION: DEVELOPMENT-SNAPSHOT-2026-03-16-a
-  AZURE_VAULT_SIGN_COMMAND: ${{ secrets.AZURE_KEY_VAULT_URI != '' && '"%USERPROFILE%\.dotnet\tools\AzureSignTool.exe" sign --azure-key-vault-url "%AZURE_KEY_VAULT_URI%" --azure-key-vault-managed-identity --azure-key-vault-certificate "%AZURE_KEY_VAULT_CERTIFICATE_NAME%" --file-digest sha256 --timestamp-rfc3161 "http://timestamp.digicert.com" --timestamp-digest sha256 --verbose' || '' }}
+  AZURE_VAULT_SIGN_COMMAND: ${{ secrets.AZURE_KEY_VAULT_URI != '' && '"%UserProfile%\.dotnet\tools\AzureSignTool.exe" sign --azure-key-vault-url "%AZURE_KEY_VAULT_URI%" --azure-key-vault-managed-identity --azure-key-vault-certificate "%AZURE_KEY_VAULT_CERTIFICATE_NAME%" --file-digest sha256 --timestamp-rfc3161 "http://timestamp.digicert.com" --timestamp-digest sha256 --verbose' || '' }}
 
 defaults:
   run:
--- a/.github/workflows/swift-toolchain.yml
+++ b/.github/workflows/swift-toolchain.yml
@@ -317,7 +317,6 @@ on:
 
 env:
   PINNED_BOOTSTRAP_TOOLCHAIN_VERSION: DEVELOPMENT-SNAPSHOT-2026-03-16-a
-  AZURE_VAULT_SIGN_COMMAND: ${{ secrets.AZURE_KEY_VAULT_URI != '' && '"%UserProfile%\.dotnet\tools\AzureSignTool.exe" sign --azure-key-vault-url "%AZURE_KEY_VAULT_URI%" --azure-key-vault-managed-identity --azure-key-vault-certificate "%AZURE_KEY_VAULT_CERTIFICATE_NAME%" --file-digest sha256 --timestamp-rfc3161 "http://timestamp.digicert.com" --timestamp-digest sha256 --verbose' || '' }}
 
 defaults:
   run:
@@ -4543,9 +4542,6 @@ jobs:
     name: Package Tools
     needs: [compilers, macros, debugging_tools, devtools, stdlib, sdk]
     runs-on: ${{ inputs.default_build_runner }}
-    env:
-      AZURE_KEY_VAULT_URI: ${{ secrets.AZURE_KEY_VAULT_URI }}
-      AZURE_KEY_VAULT_CERTIFICATE_NAME: ${{ secrets.AZURE_KEY_VAULT_CERTIFICATE_NAME }}
 
     strategy:
       fail-fast: false
@@ -4553,6 +4549,18 @@ jobs:
       matrix: ${{ fromJSON(inputs.host_matrix) }}
 
     steps:
+      - name: Select signing mode
+        id: signing_mode
+        if: inputs.signed
+        env:
+          AZURE_KEY_VAULT_URI: ${{ secrets.AZURE_KEY_VAULT_URI }}
+        run: |
+          if ($env:AZURE_KEY_VAULT_URI) {
+            echo "azure_vault=true" >> $env:GITHUB_OUTPUT
+          } else {
+            echo "azure_vault=false" >> $env:GITHUB_OUTPUT
+          }
+
       - uses: actions/checkout@v4.2.2
         with:
           ref: ${{ inputs.ci_build_revision }}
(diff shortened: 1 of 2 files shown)


.github/workflows/swift-toolchain.yml
```diff
@@ -310,15 +310,14 @@ on:
...
         required: false
-      TRUSTED_SIGNING_TEST_PROFILE:
-        required: false
-      TRUSTED_SIGNING_PROD_PROFILE:
+      AZURE_KEY_VAULT_CERTIFICATE_NAME:
         required: false
 
 env:
   PINNED_BOOTSTRAP_TOOLCHAIN_VERSION: DEVELOPMENT-SNAPSHOT-2026-03-16-a
+  AZURE_VAULT_SIGN_COMMAND: ${{ secrets.AZURE_KEY_VAULT_URI != '' && '"%USERPROFILE%\.dotnet\tools\AzureSignTool.exe" sign --azure-key-vault-url "%AZURE_KEY_VAULT_URI%" --azure-key-vault-managed-identity --azure-key-vault-certificate "%AZURE_KEY_VAULT_CERTIFICATE_NAME%" --file-digest sha256 --timestamp-rfc3161 "http://timestamp.digicert.com" --timestamp-digest sha256 --verbose' || '' }}
```
compnerd: Can we use the canonical spelling for the variables?


```suggestion
  AZURE_VAULT_SIGN_COMMAND: ${{ secrets.AZURE_KEY_VAULT_URI != '' && '"%UserProfile%\.dotnet\tools\AzureSignTool.exe" sign --azure-key-vault-url "%AZURE_KEY_VAULT_URI%" --azure-key-vault-managed-identity --azure-key-vault-certificate "%AZURE_KEY_VAULT_CERTIFICATE_NAME%" --file-digest sha256 --timestamp-rfc3161 "http://timestamp.digicert.com" --timestamp-digest sha256 --verbose' || '' }}
```

.github/workflows/swift-toolchain.yml
```diff
@@ -4770,46 +4745,28 @@ jobs:
...
+        if: inputs.signed && env.AZURE_VAULT_SIGN_COMMAND != ''
+        run: |
+          $UserPropsPath = Join-Path -Path "${{ github.workspace }}/SourceCache/swift-installer-scripts/platforms/Windows" -ChildPath "$($env:USERNAME).props"
+          @"
+          <Project>
+            <PropertyGroup>
+              <SignCommand>$env:AZURE_VAULT_SIGN_COMMAND</SignCommand>
+            </PropertyGroup>
+          </Project>
+          "@ | Out-File -FilePath $UserPropsPath -Encoding utf8
```
compnerd: I wonder if this is worth the effort - can we simply do `-p:SignCommand="..."` as a build parameter?

.github/workflows/swift-toolchain.yml
```diff
@@ -5369,46 +5299,28 @@ jobs:
...
```
compnerd: Likewise

.github/workflows/swift-toolchain.yml
```diff
@@ -5696,46 +5608,28 @@ jobs:
...
+        if: inputs.build_android && inputs.signed && env.AZURE_VAULT_SIGN_COMMAND != ''
+        run: |
+          $UserPropsPath = Join-Path -Path "${{ github.workspace }}/SourceCache/swift-installer-scripts/platforms/Windows" -ChildPath "$($env:USERNAME).props"
+          @"
+          <Project>
+            <PropertyGroup>
+              <SignCommand>$env:AZURE_VAULT_SIGN_COMMAND</SignCommand>
+            </PropertyGroup>
+          </Project>
+          "@ | Out-File -FilePath $UserPropsPath -Encoding utf8
```
compnerd: I guess here too.

.github/workflows/swift-toolchain.yml
```diff
@@ -5813,54 +5706,36 @@ jobs:
...
+        if: inputs.signed && env.AZURE_VAULT_SIGN_COMMAND != ''
+        run: |
+          $UserPropsPath = Join-Path -Path "${{ github.workspace }}/SourceCache/swift-installer-scripts/platforms/Windows" -ChildPath "$($env:USERNAME).props"
+          @"
+          <Project>
+            <PropertyGroup>
+              <SignCommand>$env:AZURE_VAULT_SIGN_COMMAND</SignCommand>
+            </PropertyGroup>
+          </Project>
+          "@ | Out-File -FilePath $UserPropsPath -Encoding utf8
```
compnerd: Seems like it would end up with even smaller change!

.github/workflows/swift-toolchain.yml
```diff
@@ -5959,54 +5833,36 @@ jobs:
...
```
compnerd: Feeling worth it!

.github/workflows/swift-toolchain.yml
```diff
@@ -6152,49 +6008,31 @@ jobs:
...
```
compnerd: Definitely

.github/workflows/swift-toolchain.yml
speednoisemovement: We'd need to restructure the swift-installer-scripts change: my understanding is `-p` makes it unoverridable, so the fallback logic can't trigger.  To not pass it at all, we'd need a conditional call site which is noisier than this.

Installer scripts change would be something like
```
-p:OverrideSignCommand="..."
// in MSBuild:
<SignCommand Condition="$(OverrideSignCommand) != ''">$(OverrideSignCommand)</SignCommand>
<SignCommand Condition="$(OverrideSignCommand) == ''">signtool.exe ...</SignCommand>
```

.github/workflows/swift-toolchain.yml
speednoisemovement: Done

Steelskin (approved):
LGTM once Saleem's concerns are addressed.

.github/workflows/swift-toolchain.yml
speednoisemovement: Changed this to local property in swift-installer-scripts which will allow overriding. Done

.github/workflows/swift-toolchain.yml
```diff
@@ -310,15 +310,14 @@ on:
...
         required: false
-      TRUSTED_SIGNING_TEST_PROFILE:
-        required: false
-      TRUSTED_SIGNING_PROD_PROFILE:
+      AZURE_KEY_VAULT_CERTIFICATE_NAME:
         required: false
 
 env:
   PINNED_BOOTSTRAP_TOOLCHAIN_VERSION: DEVELOPMENT-SNAPSHOT-2026-03-16-a
+  AZURE_VAULT_SIGN_COMMAND: ${{ secrets.AZURE_KEY_VAULT_URI != '' && '"%UserProfile%\.dotnet\tools\AzureSignTool.exe" sign --azure-key-vault-url "%AZURE_KEY_VAULT_URI%" --azure-key-vault-managed-identity --azure-key-vault-certificate "%AZURE_KEY_VAULT_CERTIFICATE_NAME%" --file-digest sha256 --timestamp-rfc3161 "http://timestamp.digicert.com" --timestamp-digest sha256 --verbose' || '' }}
```
compnerd: not sure how I feel about this being passed through the environment, so would appreciate a peek from Cory.

compnerd (approved):

.github/workflows/swift-toolchain.yml
speednoisemovement: + @coryhardman ^

coryhardman: Not concerned about it from a security standpoint. The command is a static literal in the workflow — the only dynamic piece is the emptiness check on the secret, and the URI/cert name stay as `%VAR%` for cmd to expand at runtime. No attacker-controlled input reaches it, and anyone who can edit the workflow already has arbitrary execution in this job after `azure/login`.

One thing worth checking: workflow-level `env` can be overridden by a later `$GITHUB_ENV` write in the same job, and this job runs a fair amount of external code (checked-out wixproj builds, a global dotnet tool install) before the sign step. If any of that can append to `GITHUB_ENV`, it replaces the sign command *and* flips the `if:` conditions that pick the signing path. Setting it at step level or re-deriving it right before use would close that off.

coryhardman (approved):

.github/workflows/swift-toolchain.yml
speednoisemovement: Thanks!

I restructured to derive signing mode in a step each job (so it can't be overwritten) and the sign command right when it's used

PR #1143, opened 2026-09-01
Update package_tools for bootstrapping

speednoisemovement:


.github/workflows/swift-toolchain.yml
```diff
@@ -5712,9 +5712,45 @@ jobs:
...
+        uses: thebrowsercompany/gha-download-tar-artifact@59992d91335d4ecba543c8535f7d07238e42125d # main
+        with:
+          name: Android-i686-compiler-runtime
+          path: ${{ github.workspace }}/BuildRoot/Library
+      - name: Download Android x86_64 Compiler Runtime
+        if: inputs.build_android
+        uses: thebrowsercompany/gha-download-tar-artifact@59992d91335d4ecba543c8535f7d07238e42125d # main
         with:
-          name: Windows-${{ matrix.arch }}-stdlib
+          name: Android-x86_64-compiler-runtime
```
compnerd: The android runtime bits don't make sense. We never run the package manager on Android.

speednoisemovement: I just went down a rabbit hole on this. The gist is this matches build.ps1 in a way we weren't previously: the Swift.org 6.3 contains the full matrix of runtimes; our `20260417.1` omits them.

speednoisemovement: We talked offline; removed

.github/workflows/swift-toolchain.yml
```diff
@@ -5801,23 +5798,437 @@ jobs:
...
+                    throw "'$ImagePath' has RT_MANIFEST resource #$ResourceID with UAC execution-level metadata"
+                  }
+                } finally {
+                  Remove-Item -Path $ManifestPath -Force -ErrorAction SilentlyContinue
+                }
+              }
+              Write-Host "Validated '$ImagePath' RT_MANIFEST resources: $($ManifestIDs -join ', ')"
+            }
+          }
+
```
compnerd, line 6069: I wonder if we really need the dynamism here or if we should hardcode the list. Again, 1:1 correspondence to build.ps1 is a non-goal, what we want here is the fastest, simplest correct build of the toolchain.

speednoisemovement: Done. I'm a little concerned about maintenance, but if it's a real problem we can talk about restoring this

compnerd (approved):

PR #1148, opened 2026-09-16
Update swift_test job for bootstrapping

speednoisemovement:


.github/workflows/swift-toolchain.yml
```diff
@@ -7661,6 +7672,13 @@ jobs:
           name: ${{ inputs.build_os }}-${{ inputs.build_arch }}-early-swift-driver
           path: ${{ github.workspace }}/BinaryCache/swift-driver
 
+      - name: Disable Git line ending conversion
```
speednoisemovement, line 7675: I think there must have been a GH runner change; `autocrlf` was causing a few tests to fail

compnerd (approved):

